PR #6, opened 2018-10-29
add saving of the image digest to `In`

topherbullock:
Having the digest is useful when another step needs to know which version of a registry-image-resource was pulled down by a get step.


topherbullock:
Thanks @vito ! 
you the real mvp

Issue #1, opened 2018-08-07
Basic auth support


vito:
Right now there's no way to configure auth, which means this resource is only useful for fetching public images.

There's all sorts of auth we'd need to support various registries. Let's start with the simplest one: basic auth.

marco-m:
If I understand correctly now this resource supports basic auth, for example in https://github.com/concourse/concourse/blob/e5d5b9af1ef45243fe9062a86fe7392e74cd5f6a/ci/pipelines/concourse.yml#L714

    - name: unit-image
      type: registry-image
      source:
        repository: concourse/unit
        username: ((docker.username))
        password: ((docker.password))	

is this correct ?

vito:
Right now it's only supported by `put` because I needed it for pushing to work at all. So `check` and `get` still can't use basic auth for private images.

int-tt:
Thank you for a wonderful resource!
I needed to use a private repository in my work, so I tried writing PR:)  #9

vito:
finished, tested, and documented!

thanks for the PR, @int-tt!

PR #9, opened 2018-12-14
feat:support private repository

int-tt:
#1 
not using authentication information in `in` and `check` , we responded to problems that could not be used in private repository :)

check: `ERROR failed get image digest: UNAUTHORIZED: "authentication required" `
in: `ERRO[0004] failed to extract image: UNAUTHORIZED: "authentication required" `

### DIFF
--- a/cmd/check/main.go
+++ b/cmd/check/main.go
@@ -5,7 +5,9 @@ import (
 	"os"
 
 	resource "github.com/concourse/registry-image-resource"
+	"github.com/google/go-containerregistry/pkg/authn"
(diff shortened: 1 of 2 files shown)


vito:
thanks! this looks great. :slightly_smiling_face:

i was waiting to do this until i had a way of writing tests for it, but i'll pull your changes in locally and see if i can write up some tests myself, since that'll take some additional work that would be difficult for someone else to do (i.e. setting up a private docker hub repository for testing).

once i have everything in order i'll push this + the tests. :+1:

PR #12, opened 2019-01-27
adds repo & tag to metadata for in & out cmd

osis:
Returning this metadata will help identify targeted repos and tags on the resource details view.


osis:
Grabbed the PR image and put it in my pipeline. 

screen shot 2019-01-27 at 12 33 58 am

osis:
Makes sense to add it to out as well. I'll adjust!

Issue #7, opened 2018-11-27
Question about the plan for tagging built images


pn-santos:
_Disclaimer: Not sure if this is the right place to post this, or if it's "too soon" in the development for asking (seeing this is still "beta"). Apologies if it is._

I was wondering what the vision/plan is for tagging images. I know put is still very early stages but this resource as it stands, requires tags to be specified "statically" (in the pipeline config). However I think it's a frequent use case to tag built images with runtime generated values (like auto-incremented build nr, or git rev, etc.). 

I've been experimenting with this resource (and the [builder task](https://github.com/concourse/builder/) - they work well and much faster than the docker-image resource btw :+1:) and I've been pondering whether it would be useful to add the ability to read tags from a file when specifying a put (for example). But not sure if it fits in the plans for this resource.

From a design perspective it seems to go against specifying the tag in the source definition, since that implies the resource is tracking a specific tag, so for symmetry the put should update the tag...

vito:
Good question (sorry for the delay, end-of-year vacations took me away for a while)! Tags were a major source of confusion in the Docker Image resource and the interface around them was constantly in flux. It was scary enough that I haven't added support for it yet, but if we can all agree on the behavior I'd be happy to merge a PR for it.

And thanks for the feedback re: the `builder` task, glad to hear it's working out well! :+1: 

For now, I think it still makes sense to have a tag in `source` since there needs to be some way of specifying where the versions are coming from. This may change in the future with concourse/rfcs#1 (more on that later).

I think we could pretty much replicate the `additional_tags` feature from the Docker Image resource, and ignore the other tag-related params. It would push the tags read from the specified file (whitespace/linebreak-separated), in addition to the tag specified by `source` (or `latest` if not specified, so its default is consistent with `get` and `check`).

There's usually a central tag to update with `put`, and usually people want to add *additional* tags, e.g. push to `latest` or `dev` while also tagging it with specific versions. Something like `latest`, `1.2.3`, `1.2`, `1`.

[...]

PR #13, opened 2019-02-01
Add support for additional_tags put param

pn-santos:
Hi @vito tried to implement what was discussed in https://github.com/concourse/registry-image-resource/issues/7. I'm a complete n00b with golang so reviewer beware :warning: 

I tried looking in go-containerregistry to see if there was an easier way to just push new manifests, but they [don't seem to support that yet](https://github.com/google/go-containerregistry/issues/349) so I ended up just pushing the image with different tags (the go-containerregistry will still go through all layers but will end up skipping them all because they already exist in the repo).

I did change `tag` property to `tags` in the out metadata however I think that should not break any existing use of the resource (was the only thing that changed the existing API).

I also need a little guidance on how I should go about writing the tests... do you need to setup a private docker registry from them to work? I would assume the best place for testing this would be `out_test.go` ?

I did a very quick test with a pipeline and it worked (first push uploads all layers, subsequent pushes are almost instant and only push the manifest).

### DIFF
--- a/out_test.go
+++ b/out_test.go
@@ -62,7 +62,7 @@ var _ = Describe("Out", func() {
 		Expect(err).ToNot(HaveOccurred())
 	})
 
-	Describe("pushing an OCI image tarball", func() {
+	Context("pushing an OCI image tarball", func() {
 		var randomImage v1.Image
 
 		BeforeEach(func() {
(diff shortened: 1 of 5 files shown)


out_test.go
```diff
@@ -9,14 +9,14 @@ import (
...
-	"github.com/google/go-containerregistry/pkg/v1"
+	v1 "github.com/google/go-containerregistry/pkg/v1"
 	"github.com/google/go-containerregistry/pkg/v1/random"
 	"github.com/google/go-containerregistry/pkg/v1/remote"
 	"github.com/google/go-containerregistry/pkg/v1/tarball"
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
 
-	"github.com/concourse/registry-image-resource"
+	resource "github.com/concourse/registry-image-resource"
```
pn-santos: not really sure why but vscode insisted on adding these alias on save, not sure if needed or something with my config

vito:
Yo, thanks for taking this on! The changes look great so far, only thing left is tests as you said. Pushing the same manifest makes sense, I think that's what Docker does anyway. Or at least that's what the `docker-image` resource used to do.

To test with `out_test.go` you currently need to configure these values via env vars:

https://github.com/concourse/registry-image-resource/blob/d34b345b1d57d4a0ee6b2dbda93dd91c60ac399d/suite_test.go#L31-L39

I think you could just add another test that configures `additional_tags` and verifies that each tag was also pushed and points to the same digest.

Maybe something like this?:

```go
Context("with additional_tags", func() {
	BeforeEach(func() {
		// set up additional_tags to point to file containing "some\ntags"
	})

	It("pushes with each tag in addition to the tag in 'source'", func() {
		randomDigest, err := randomImage.Digest()
		Expect(err).ToNot(HaveOccurred())

		for _, tag := range []string{"latest", "some", "tags"} {
			name, err := name.ParseReference(req.Source.Repository+":"tag, name.WeakValidation)
			Expect(err).ToNot(HaveOccurred())

			image, err := remote.Image(name)
			Expect(err).ToNot(HaveOccurred())

			pushedDigest, err := image.Digest()
			Expect(err).ToNot(HaveOccurred())

			Expect(pushedDigest).To(Equal(randomDigest))
		}
	})
})
```

pn-santos:
@vito finally was able to get some time to get this done. Hope everything checks out (btw I included a commit with the recently added ARGs - which I think you might have missed in https://github.com/concourse/registry-image-resource/commit/53579a51ee15abf7c08faae3d6a6e1032886fed7)

PR #34, opened 2019-04-17
Make sure check won't fail if image tag not exist on registry server.

### LINKED ISSUE #32: Proposal: don't let "check" step failed if `source.tag` not exist on registry server
In current implement, even you don't specify `source.tag`, `put` step still push the image to the default `latest` tag. 
Our user wants to avoid using `latest` tag and only keep versioned tag. In this case, registry-image `check` step will fail. Could you avoid this kind of failure in `registry-image-resource`?(the same case won't fail in `docker-image-resource`)

SimonXming:
For fix #32, the PR make sure `check` step won't fail if the image tag does not exist on the registry server.

### DIFF
--- a/check_test.go
+++ b/check_test.go
@@ -210,4 +210,36 @@ var _ = Describe("Check", func() {
 			})
 		})
 	})
+
+	Context("when invoked with not exist image", func() {
+		BeforeEach(func() {
+			req.Source = resource.Source{
+				Repository: "concourse/test-image-static",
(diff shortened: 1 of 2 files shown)


SimonXming:
Hi @vito, any updates about reviewing? This is a blocking issue for us, Thanks in advance.😄

SimonXming:
> Looks good, sorry for the wait! Would you mind adding a test though? (Thanks!)

@vito Sure, Test case added.😄

PR #36, opened 2019-04-27
Allow configuration of `platform`

cirocosta:
Hey,

I was giving a try at running Concourse on ARMv7 (worked!), and one of the modifications needed was making `registry-image` gather container images built for such a platform.

Without the explicit use of `platform`, the underlying library ends up
using its own default (amd64). This is a problem for those running the
resource in other architectures.

To achieve that:

- bumps `go-containerregistry` to make
- sets `Platform` to the `{GOOS, GOARCH}` by default, allowing the
  possibility of having that set through a `platform` field in the
  resource config.

Example:

```yaml
resources:
- type: registry-image
  name: busybox-arm
  source:
    repository: busybox
    platform: {os: linux, architecture: arm}
```

The biggest uncertainty that I have regarding this at the moment is the default values - should we leave them as the values chosen for GOOS and GOARCH at build time, or make it by default a specific setting (like, `linux amd64`?) 

Also, regarding testing, should we publish an image that is arm-based to DockerHub so we can ensure that in the tests we gather the expected digest?

Wdyt?

Thanks!


vito:
Defaulting based on `GOOS`/`GOARCH` sounds sensible. So we would build + `docker push` + release a `linux-arm` variant of this resource, and anyone using it would have it default to the appropriate platform. As long as it's overrideable, it seems less arbitrary than defaulting to `linux-amd64` everywhere.

The only risk would be whether folks would find this surprising, e.g. if they're fetching the image but not intending to actually run it on the same platform. But I think the trade-off in brevity when using the resource for the "common case" (`image_resource:`/`resource_types:`) on other platforms makes it worth the risk.

vito:
> Also, regarding testing, should we publish an image that is arm-based to DockerHub so we can ensure that in the tests we gather the expected digest?

Hmm I guess so. This would be our first time supporting other platforms in a base resource type, but if we're gonna start somewhere this resource makes the most sense. Would this complicate the test setup?

Infinoid:
I needed this, but the PR no longer applies cleanly.

I cooked up a patch based loosely on this, which seems to work.  The patch is at Infinoid@39276ce if it helps.

natto1784:
also waiting for this

suite_test.go
```diff
@@ -42,7 +42,7 @@ const OLDER_LIBRARY_DIGEST = "sha256:2131f09e4044327fd101ca1fd4043e6f3ad921ae7ee
 
 // see testdata/static/Dockerfile
 const OLDER_STATIC_DIGEST = "sha256:7dabedca9d367a71d1cd646bd8d79f14de7b07327e4417ab691f5f13be5647a9"
-const LATEST_STATIC_DIGEST = "sha256:29eddd288c312215a0af1070d26478b1f530a3137d4589314df1bc79e586860a"
+const LATEST_STATIC_DIGEST = "sha256:6d89d782e9c924098af48daa6af692d14aba9f4a92338ea04603d99ef68395df"
```
xtremerui, line 45: This is changed when running
```
docker buildx build --platform linux/arm64/v8,linux/amd64 -t concourse/test-image-static:latest --push .
```
for building multi-arch image.

natto1784:
NO WAY LETS GOOOOO

natto1784:
this is the best thing that has happened today and i jlust installed woodpecker CI, concourse im coming back

Issue #41, opened 2019-05-28
Feature request: Support content trust while image pushing?


SimonXming:
## What challenge are you facing?
There are requests from users to add content trust(similar function like `DOCKER_CONTENT_TRUST` and `DOCKER_CONTENT_TRUST_SERVER`) in registry-image-resource. So they could sign image with their Notary server in concourse pipeline.

## What would make this better?
Add content trust feature at resource level or library level. I noticed registry-resource is using `github.com/google/go-containerregistry` and it doesn't support content trust feature right now. I have created https://github.com/google/go-containerregistry/issues/461, not sure if this feature request will be accepted. 

## Are you interested in implementing this yourself?
Yes, but I should wait for https://github.com/google/go-containerregistry/issues/461.

vito:
Interesting, let's keep an eye on the upstream issue for now. Sounds reasonable to me though I don't have a whole lot of experience with it. I'm assuming the change wouldn't be too invasive.

SimonXming:
Hi @vito, `go-containerregistry` developer suggest implement it at resource-type level([Link](https://github.com/google/go-containerregistry/issues/461#issuecomment-497028546)). So I would start working on the code change recently. WDYT?

SimonXming:
Hi @vito, Is that OK adding this feature?

vito:
@SimonXming Sorry but I'm in a similar position to the folks upstream, not being familiar enough with it to say yes or no. For now I would suggest just forking and making the change and then I'll have a better idea of what's involved, whether it should remain a fork, or whether there's another approach we could take. Does that sound OK?

SimonXming:
Hi @vito, could you review PR #43 when it's convenient for you? Thanks

pivotal-jamie-klassen:
Fixed in #46

PR #43, opened 2019-06-13
WIP: Add support for content trust.

SimonXming:
Add image sign feature in `put` step. #41


SimonXming:
The current design is 
```
  - put: some-docker-image
    params:
      image: image/image.tar
      content_trust:
        enable: true
        server: https://127.0.0.1:4443
        config_dir: notary-config-dir
        repository_passphrase: "xxxxxxx"
```
Some user suggest support both having `config_dir` or providing values directly.
```
  - put: some-docker-image
    params:
      image: image/image.tar
      content_trust:
        enable: true
        server: https://127.0.0.1:4443
        server_tls_cert: "xxxxxx"
        server_tls_key: "xxxxxx"
        signing_key: "xxxxxx"
        repository_passphrase: "xxxxxxx"
```
@vito what do you think from `registry-image-resource` perceptive? Any comments?

SimonXming:
The current design is 
```
  - put: some-docker-image
    params:
      image: image/image.tar
      content_trust:
        enable: true
        server: https://127.0.0.1:4443
        config_dir: notary-config-dir
        repository_passphrase: "xxxxxxxx"
```
Some user suggest support both having `config_dir` or providing values directly.
```
  - put: some-docker-image
    params:
      image: image/image.tar
      content_trust:
        enable: true
        server: https://127.0.0.1:4443
        server_tls_cert: "xxxxxxxxx"
        server_tls_key: "xxxxxxxxx"
        config_dir: notary-config-dir
        signing_key: "xxxxxxxxx"
        repository_passphrase: "xxxxxxxx"
```
@vito What do you think from `registry-resource-type` perspective? Any comments?

SimonXming:
The current design is 
```
  - put: some-docker-image
    params:
      image: image/image.tar
      content_trust:
        enable: true
        server: https://127.0.0.1:4443
        config_dir: notary-config-dir
        repository_passphrase: "xxxxxxx"
```
Some user suggest support both having `config_dir` or providing values directly
```
  - put: some-docker-image
    params:
      image: image/image.tar
      content_trust:
        enable: true
        server: https://127.0.0.1:4443
        server_tls_cert: "xxxxxxx"
        server_tls_key: "xxxxxxx"
        config_dir: notary-config-dir
        signing_key: "xxxxxxx"
        repository_passphrase: "xxxxxxx"
```
@vito what do you think from `registry-resource-type` perspective? Any comments?

vito:
Hmm I'm a little worried that there's so much code involved with this, and I wouldn't really have the context to support and maintain it. One of the problems with the `docker-image` resource was that it kept growing and growing, so I'm just being a little cautious here.

Would it be possible to implement this as a separate resource? It looks like it performs another set of side-effects using data produced by the image. Could it be done in a separate `put` step that runs after the `registry-image` resource and only deals with the content trust aspect?

SimonXming:
Ideally, since content trust actually push some data(with tag) into a notary server. It make sense to use a separate resource to track.
But, let say a user have both registry and notary, he only want deal with signed tags(for pull/push).

1. For separate resource, check following example. IMO, it looks pretty confusing
```
resources:
- name: dest-image
  type: registry-image
  source:
    repository: dest-image-path

- name: signing-dest-image
  type: signing-image
  source:
    repository: dest-image-path
    content_trust: {....}

jobs:
- name: separate-resource-example-1
  plan:
  - get: signing-dest-image # if signing-image don't retrieve tarball
    trigger: true
  - get: dest-image         # then we need to use `dest-image` to retrieve tarball
  - task: some-task
  - put: dest-image         # upload image/image.tar
    params:
      image: image/image.tar
  - put: signing-dest-image # publish extra data to notary server about image/image.tar
    params:
      image: image/image.tar
      content_trust: {....}

jobs:
- name: separate-resource-example-2
  plan:
  - get: signing-dest-image # if signing-image does retrieve tarball
    trigger: true
  # - get: dest-image         # then we don't need to this step.
                              # But if `signing-image` could pull image tarball, it should also support `push` it registry-image
  - task: some-task
  - put: dest-image         # upload image/image.tar
    params:
      image: image/image.tar
  - put: signing-dest-image # publish extra data to notary server
    params:
      image: image/image.tar
      content_trust: {....}
```
2. If implement as registry-image advance feature, the usage would be very straightforward
```
resources:
- name: unify-dest-image
  type: registry-image
  source:
    repository: dest-image-path
    content_trust: {....}   # enable content_trust for both pull/push

jobs:
- name: unify-resource-example-1
  plan:
  - get: unify-dest-image   # if signing-image don't retrieve tarball
    trigger: true
  - task: some-task
  - get: signing-dest-image #
    params:
      image: image/image.tar

jobs:
- name: unify-resource-example-2
  plan:
  - get: unify-dest-image   # if signing-image don't retrieve tarball
    param:
      disable_content_trust: true # disable content_trust for only for pull
    trigger: true
  - task: some-task
  - get: signing-dest-image
    params:
      image: image/image.tar
```

In summary, my concern are:
1. Semantically, trust tags are subset of ordinary tags, it make sense to provide as advanced feature of registry-resource
2. Using separate resource will cause pretty confusing pipeline YAML
3. If use it as advanced feature, the configuration would be easy, only need define `resource.source.content_trust`
4. If using separate resource, there are two separate version history for `all version` and `signed version`, but user only care `signed version`

vito:
You're right, it's definitely a lot uglier to split it out.

I guess I'm just concerned that we had to write so much of this code ourselves. Is there no sufficiently high-level library to do all this? Or is the code here actually somewhat domain-specific?

In the long term, this could be an interesting case to point to for composability of resources - i.e. a sidecar for augmenting `check`/`get`/`put`. Definitely out of scope for this PR though. Maybe something to think about in concourse/rfcs#21.

SimonXming:
I don't think there are high-level library integrate `github.com/theupdateframework/notary` and `github.com/google/go-containerregistry`. 
Maybe there should be one, since there just too many code for `registry-image-resource` and `github.com/google/go-containerregistry` suggest we could do it on high level(https://github.com/google/go-containerregistry/issues/461)
We will evaluate the effort for doing it.

SimonXming:
Close it in favor of #46.

PR #46, opened 2019-07-08
Add support for signing image.

SimonXming:
Add image sign feature in put step. #41


SimonXming:
Close it in favor of #46.

SimonXming:
Hi @vito, this is a replacement PR for #43, please take a look at your convenience. Thanks :)

types_test.go
```diff
@@ -61,6 +61,6 @@ var _ = Describe("Source", func() {
 		json, err := json.Marshal(source)
 		Expect(err).ToNot(HaveOccurred())
 
-		Expect(json).To(MatchJSON(`{"repository":"foo","tag":"0"}`))
+		Expect(json).To(MatchJSON(`{"repository":"foo","tag":"0","content_trust":{"enable":false,"server":"","repository_key_id":"","repository_key":"","repository_passphrase":"","tls_key":"","tls_cert":""}}`))
```
vito: Would be nice to avoid this; I think if you sprinkle `,omitempty` on all the fields of the content trust struct it'll go away. Or maybe make it a pointer?

cmd/out/main.go
```diff
@@ -117,6 +118,26 @@ func main() {
 
 	logrus.Info("pushed")
 
+	if req.Source.ContentTrust.Enable {
```
vito: Rather than having an `enable` field maybe we can make `ContentTrust` a pointer and just check if it's non-nil. If a user configures any of those fields they probably want it enabled.

Although, one reason to have an explicit `Enable` field is if all the other fields are optional - in which case the only thing to do would be to enable it. Is that possible? It seems like the server field is required at least.

README.md
```diff
@@ -38,6 +38,14 @@ differences:
 * `debug`: *Optional. Default `false`.* If set, progress bars will be disabled
   and debugging output will be printed instead.
 
+* `content_trust`: *Optional.* Configuration about content trust.
```
vito: Might want to individually note which of these sub-fields are required. I'm guessing the TLS stuff for example is optional?

cmd/out/main.go
```diff
@@ -117,6 +118,26 @@ func main() {
...
+		notaryConfigDir, err := req.Source.ContentTrust.PrepareConfigDir(src)
+		if err != nil {
+			logrus.Errorf("failed to prepare notary-config-dir: %s", err)
+			os.Exit(1)
+			return
+		}
+		os.Setenv("NOTARY_CONFIG_DIR", notaryConfigDir)
```
vito: This seems kinda odd; could it be passed as an argument instead? :thinking:

types.go
```diff
@@ -14,12 +16,64 @@ type Source struct {
...
+├── gcr-config.json
+└── trust
+	└── private
+		└── <private-key-id>.key
+└── tls
+	└── <notary-host>
+		├── client.cert
+		└── client.key
+*/
+func (ct *ContentTrust) PrepareConfigDir(src string) (string, error) {
```
vito: This also seems a little odd. Why does it have to be written to disk? Is this for integrating with an external tool, or could it be changed to just pass a config argument to your library? :thinking:

SimonXming: Yes, many API from the official `github.com/theupdateframework/notary` rely on a base directory to store where all the trust files. For example, https://github.com/theupdateframework/notary/blob/28e8e6d5c2221984e9acd1f2bab017cfdcdf106c/client/client.go#L60.

cmd/out/main.go
SimonXming: Since `github.com/theupdateframework/notary` rely on a `trust_directory`. In order to keep consistency, I'm using `NOTARY_CONFIG_DIR` to explicitly specified the directory.

types_test.go
SimonXming: Thanks, I will try to make it a pointer. 😄

cmd/out/main.go
SimonXming: Yeah, `server` field is optional because there are default Notary server `https://notary.docker.io`. But AFAIK, at least `repository_key_id` and `repository_key` are not optional. So remove `enable` field sounds good choice.

README.md
SimonXming: Sure, will update doc.

cmd/out/main.go
vito: I just meant that it felt odd to have to set an env var in order to use a Go API. It's kind of a mix of system-level config and language-level config. Could this be pushed under the hood and handled by your library? Where does this env var get used? I can't seem to find it anywhere.

types.go
```diff
@@ -14,12 +16,64 @@ type Source struct {
...
+	configDir := filepath.Join(src, ".notary")
+	os.Mkdir(configDir, os.ModePerm)
```
vito: Missing `err` check here and through the rest of the method. There are a few assignments that are unchecked as well.

cmd/out/main.go
vito: It looks like it gets parsed out in [`gcr.NewTrustedGcrRepository`](https://github.com/simonshyu/notary-gcr/blob/8d81a216e1287b6e8d51078c3b07da0ecfb1d7e7/pkg/gcr/gcr.go#L18-L25). The config gets written to disk and then loaded up immediately after within the same Go process, which is what feels weird.

Does something else use this directory format? If not, couldn't the config just be passed directly to the API?

Does anything else use that same env var? If not, couldn't it just be passed as an argument?

Thanks!

SimonXming: OK, I will passed it as an argument.

types.go
SimonXming: Thanks. :)

SimonXming:
Hi @vito, please take a look at your convenience. Thanks :)

Issue #48, opened 2019-07-22
Ability to use semver version for tag


roshankadasani:
When using `additional_tags` option in the `put` params, the resource tags and pushes the images with the ones specified in the additional tags file and also the `tag` specified in resources' source. I am using [semver concourse resource](https://github.com/concourse/semver-resource) for versioning tags and I want to push just the tag specified in the file. Or is there a way I can pass my servem verion to the `tag` in the sources?

Resource
```yaml
- name: docker-image
  type: registry-image
  source:
    repository: {{REPO_NAME}}
    username: ((username))
    password: ((password))
    tag: ##<servem-verion-file>##
```

Job
```yaml
- put: docker-image
    params: 
      image: image/image.tar
      additional_tags: version/number
```

If passing the servem version to the `tag` is not possible can we use something similar to `tag_file` in the [docker image resource](https://github.com/concourse/docker-image-resource#out-push-an-image-or-build-and-push-a-dockerfile)?

vito:
There's some pre-existing discussion on this here: https://github.com/concourse/registry-image-resource/issues/44

I wonder if we really need two discrete versioning schemes: one that is entirely tag-based with semver or something for ordering (and digests annotating them to ensure the versions are precise, since tags move), and one that is entirely digest-based, tracking a single tag.

sh-srees:
> one that is entirely tag-based with semver or something for ordering

In working to generate reusable pipelines for multiple projects, not being able to specify the tag to be shipped to a dockerhub makes using the registry-image resource not very useful. 

I cant have a task build an image, tag it date+gitsha, and upload it. Versioning is tracked separately as a file and is really what is moving through the pipeline. But since I can put additional_tags but cannot get additional_tags, I'm a bit stuck.

I'd also like to mention theres an oddity here, in that docker doesn't treat digests inside a tag as a "version" in the same way other resources do. For example in git, each version is a separate thing that you can pull and inspect and move through the pipeline. In docker, once you've overwritten a tag to point to a certain digest, its not like you can go back to the previous digest. A tag is essentially a single version, unchangeable. For a lot of companies, tags are used such that they are write once, not modifiable, so we generate a lot of tags and other systems increment those tag pointers on various systems.

mattburgess:
[...]

PR #51, opened 2019-08-02
Minor optimization: skip pushing additional tag if the same as source tag

pn-santos:
Ignore additional tag value if it happens to be the same as the source tag (to avoid unnecessary push).

I also had to "fix" the Dockerfiles, since it it was looking for the assets in `src/assets` when you use buildkit (not sure why but it works fine if docker is used so not sure if it's a bug of docker/buildkit - but with the absolute path it works with both).

On a more curious note, using docker `19.03.1` with buildkit enabled the integration tests aren't run (not really sure why). After I disabled buildkit and ran the exact same docker command the tests were run.

I did a small change to the IT to ensure the skip part doesn't affect the overall functionality but not sure if it is possible to actually check if a push was actually made or not...

### DIFF
--- a/dockerfiles/alpine/Dockerfile
+++ b/dockerfiles/alpine/Dockerfile
@@ -12,7 +12,7 @@ RUN set -e; for pkg in $(go list ./...); do \
 
 FROM alpine:edge AS resource
 RUN apk add --no-cache bash tzdata ca-certificates unzip zip gzip tar
-COPY --from=builder assets/ /opt/resource/
+COPY --from=builder /assets/ /opt/resource/
 RUN chmod +x /opt/resource/*
 
 FROM resource AS tests
--- a/cmd/out/main.go
+++ b/cmd/out/main.go
@@ -73,6 +73,11 @@ func main() {
 
 	var extraRefs []name.Reference
 	for _, tag := range tags {
+		if tag == req.Source.Tag() {
+			logrus.Debugf("ignoring additional tag %q (same as source tag)", tag)
+			continue
+		}
+
 		n := fmt.Sprintf("%s:%s", req.Source.Repository, tag)
 
 		extraRef, err := name.ParseReference(n, name.WeakValidation)
--- a/dockerfiles/ubuntu/Dockerfile
+++ b/dockerfiles/ubuntu/Dockerfile
@@ -18,7 +18,7 @@ RUN apt-get update \
         unzip \
         zip \
       && rm -rf /var/lib/apt/lists/*
-COPY --from=builder assets/ /opt/resource/
+COPY --from=builder /assets/ /opt/resource/
 RUN chmod +x /opt/resource/*
 
 FROM resource AS tests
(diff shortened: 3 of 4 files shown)


vito:
The optimization itself makes sense, though I wonder how this came about. I don't expect the additional push would actually take that long, since everything is already pushed. So I'm not sure if it's worth the additional check.

pn-santos:
It's not critical, it just bothered me when I saw the logs and it was going through all the layers for the same tag.

In my use case I'm generating a list of tags and one of the values happens to be the same as the one tracked by the source. I can't really check from the task script what the tag the source is tracking, so I thought it made sense that being "additional" tags to have the resource itself skip a tag if it's the same as the one tracked by the source (since that is always pushed anyway).

That and the fact that it took me 5 min to make the PR changes I though I'd put it up for consideration ^_^.

But it's like you say: all that it saves is a couple of API calls, nothing is actually uploaded. If it's more valuable to keep the logic simple (I understand the value of minimizing "exceptional" behavior) you'll get no argument here. Like I said this wasn't a huge effort investment.

vito:
Yeah I'm mainly valuing the simplicity of implementation here, as you pointed out. :slightly_smiling_face: Thanks for taking the time to submit a PR for it, but I think I'll just close this for now. Glad it didn't take up too much of your time. :sweat_smile: 

I still think there's a kind of weird relationship between the source tag and the additional tags, but I haven't had the time to dig deeper. I guess while I'm thinking about it, what were the kinds of tags you were pushing? It may be more evidence in support of https://github.com/concourse/registry-image-resource/issues/48#issuecomment-517318523 :thinking:

vito:
(Actually I'm gonna re-open so I don't lose track of the conversation.)

pn-santos:
Yes it was very much in line with https://github.com/concourse/registry-image-resource/issues/48#issuecomment-517318523 (pushing `M`, `M.m` and `M.m.p` and `latest`). 

In my particular there is a "build args" file (kept in the repo) which contains a version. Sometimes that version is a semver but not always (sometimes it's a completely custom one not really under our control). There is a task that parses that version in that build args file and creates a "tags" file later used for the additional tags (if it finds a semver format it does the `M`, `M.m` and `M.m.p` tags otherwise it just uses whatever version value it finds as a single additional tag value).

The issue I was having was that one of the build args had the version `latest` which basically caused the unnecessary push (all image resources are tracking `latest`)

dockerfiles/alpine/Dockerfile
```diff
@@ -12,7 +12,7 @@ RUN set -e; for pkg in $(go list ./...); do \
 
 FROM alpine:edge AS resource
 RUN apk add --no-cache bash tzdata ca-certificates unzip zip gzip tar
-COPY --from=builder assets/ /opt/resource/
+COPY --from=builder /assets/ /opt/resource/
```
vito: quick note: buildkit has fixed the compatibility issue here via https://github.com/moby/buildkit/pull/1125 - though probably better to stick with absolute paths anyway

vito:
I'm gonna close this out - we started digging into a plan for changing how this resource versions images, which I think would eliminate the need for this. This is covered in https://github.com/concourse/registry-image-resource/issues/151 - would appreciate feedback there, and thanks for submitting this in the first place!

PR #60, opened 2019-09-05
Allow globs as part of image param for `out`

kkallday:


kkallday:
This PR needs some work. Will re-open soon.

Issue #56, opened 2019-08-27
Failed to pull specific image from Docker Hub


Zerpet:
We found an issue with registry image where an image available in Docker Hub and `docker pull`able failed in the `get` step of a job.

The image failing in the `get` step is from RabbitMQ Docker Hub:

```
rabbitmq@sha256:7c592e6c6b3510fa4353fa078ee3d5bd0b0ea5500ddee31607b964f02d0b46e2
```

This is how our resource definition and job is configured:

```
- name: rabbitmq-release
  type: registry-image
  source:
    repository: rabbitmq
    tag: 3.8-rc-management
  version: {digest: "sha256:7c592e6c6b3510fa4353fa078ee3d5bd0b0ea5500ddee31607b964f02d0b46e2"}

- name: some-job
  plan:
  - get: rabbitmq-release
    params:
      format: oci
[...]
```

In the `get` step, we get the following error message:

```
fetching rabbitmq@sha256:7c592e6c6b3510fa4353fa078ee3d5bd0b0ea5500ddee31607b964f02d0b46e2
ERRO[0000] failed to write OCI image: unsupported status code 404; body: 404 page not found 
```

However, we are able to `docker pull` this image. Worth mentioning, this image has another SHA `rabbitmq@sha256:76346c31d3eee20f82ebc2573cb0be891e4827b35e875f53cc93c9df76143732` that is `docker pull`able and it works fine with the registry image resource.

hbd:
I'm seeing this issue as well for the following images:
```fetching eclipse-mosquitto@sha256:d6d255052b779cdc704418710d30638bf387e6d8e4e707feb6a7a580e184b748
ERRO[0000] failed to write OCI image: unsupported status code 404; body: 404 page not found 
```

```fetching postgres@sha256:e874e4ecbc8f9846027a4132f32a41700e70ac6e0f804ad26fec5b3ab2cd6cff
ERRO[0000] failed to write OCI image: unsupported status code 404; body: 404 page not found 
```

```fetching zookeeper@sha256:4879178a575d76d5720602f81107be4d165107ab03de0b26669605a5d39d656d
ERRO[0000] failed to write OCI image: unsupported status code 404; body: 404 page not found 
```

PR #52, opened 2019-08-09
update dependencies

cirocosta:
it's been a while since the last time that `go-containerregistry` (and
other dependencies) have been updated.

it turns out that some interfaces changed a bit, requiring us to update
our code too.

(sending this as a preliminary PR for https://github.com/concourse/registry-image-resource/issues/49 - the `layout` package was added recently).

PR #67, opened 2019-10-21
Upgrade github.com/google/go-containerregistry to latest to fix 404s

hbd:
This fixes issue #56.


hbd:
I'm having a hard time understanding the test failure output. Could someone more familiar with the tests please advise? cc @vito @osis

hbd:
This looks similar to #52, but with an up to date `go.mod` (#52 has conflicts).

vito:
> I'm having a hard time understanding the test failure output. Could someone more familiar with the tests please advise? cc @vito @osis

Something about the new code is probably raising errors differently than how we used to check for them:

https://github.com/concourse/registry-image-resource/blob/28e8085a68dc558af294006441f766b6eb731de5/cmd/check/main.go#L122-L133

vito:
Actually, maybe not, it's actually failing before it even gets there:

https://github.com/concourse/registry-image-resource/blob/28e8085a68dc558af294006441f766b6eb731de5/cmd/check/main.go#L60

vito:
I think we just get that error earlier than we used to, actually. It looks like `remote.Image` actually tries to fetch the manifest now, whereas before that was deferred to when we call `.Digest` on the image.

hbd:
I have moved the manifest check to happen after retrieving the image instead of after the Digest (as the image will not exist if the manifest does not exist). This should fix any outstanding problems.

cmd/in/main.go
```diff
@@ -170,14 +170,8 @@ func rootfsFormat(dest string, req InRequest, image v1.Image) {
 	}
 
 	env := cfg.Config.Env
-	if len(env) == 0 {
```
vito: What was the cue for removing these? :thinking: 

It looks like the tests pass either way, but they also passed before, so I wonder whether this was important and if we need test coverage for it or if I was just kind of guessing that it'd be useful.

hbd: `config.ContainerConfig` was removed in [this commit](https://github.com/google/go-containerregistry/commit/2be26e624e32c4b3de62287e153e5e382a79b7c5), and moved to `/pkg/legacy/config. LayerConfigFile`. I didn't find a reference to `LayerConfigFile` here, so I removed the reference to `ContainerConfig` altogether.

vito: Good enough for me. If someone complains we can just introduce legacy compatibility and add a test. :+1:

Issue #68, opened 2019-10-25
Retry on 429 errors


vito:
Seems like we're getting this error a lot from Docker Hub all of a sudden:

```
ERRO[0002] failed to get remote image: unrecognized HTTP status: 429 Too Many Requests
```

We should retry with exponential backoff.

pivotal-jamie-klassen:
Just spent about half an hour investigating the case when we see a 429 in a registry-image `get` (when [fetching a layer](https://ci.concourse-ci.org/teams/main/pipelines/concourse/jobs/rc/builds/588#L5db373a0:23)), and we think we traced
https://github.com/concourse/registry-image-resource/blob/master/cmd/in/main.go#L101
-> https://github.com/concourse/registry-image-resource/blob/master/cmd/in/unpack.go#L64
-> https://github.com/google/go-containerregistry/blob/master/pkg/v1/remote/descriptor.go#L293,
(assuming we were looking at the correct implementation of `v1.Layer`) which suggests that we should see a [`github.com/google/go-containerregistry/pkg/v1/remote/transport.Error`](https://github.com/google/go-containerregistry/blob/master/pkg/v1/remote/transport/error.go#L25-L33) with a `.StatusCode` field of `429`. We're going to try to create a reproducible for the resource checking path, since that is probably a faster feedback signal (assuming that checking does something lightweight like querying for some JSON or downloading a manifest, rather than an entire layer blob which sounds kinda slow).

pivotal-jamie-klassen:
looks like it's the same story for checking, we're definitely passing into [this block](https://github.com/concourse/registry-image-resource/blob/master/cmd/check/main.go#L127) and `.StatusCode` is 429.

PR #69, opened 2019-10-28
retry on HTTP 429

pivotal-jamie-klassen:
Fixes concourse/registry-image-resource#68

Using hashicorp/go-retryablehttp instead of concourse/retryhttp was super easy! However, they have slightly different defaults: before we were using an exponential backoff starting at 1s intervals maxing at 16s intervals and setting max elapsed time of 10m. In our basic testing we just used hashicorp's defaults: 1s initial interval, 30s max interval and max 4 retries. On paper this sounds less tolerant than what we had, but tests still seem to be passing.

We wrote our own RoundTripper implementation because hashicorp has not yet done so themselves: hashicorp/go-retryablehttp#37

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 testdata/.push-output
 .envrc
+tags
--- a/cmd/check/main.go
+++ b/cmd/check/main.go
@@ -49,7 +49,7 @@ func main() {
 	}
 
 	imageOpts := []remote.Option{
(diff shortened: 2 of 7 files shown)


pivotal-jamie-klassen:
@vito's comment is especially of interest to those folks that might want to use hashicorp's library instead of our own when making requests to baggageclaim, i.e. @concourse/runtime.

PR #79, opened 2019-11-25
replace retryable transport with outer retry loop

vito:
Retrying at the transport layer actually introduced a ton of flakiness and weird behavior. `go-containerregistry` now defaults to retrying on transient errors and seems to be a lot more resilient without our own retryable transport.

Added an outer retry loop to handle 429 errors specifically via exponential backoff, starting at 5 seconds and capping at one hour. Added tests for all 3 actions to cover 429 errors returned by each registry API endpoint we hit.

Along the way, I changed the notary configuration to just use a tempdir instead of mutating the input directory. This seems a bit cleaner and just made it easier to move the code around.


out_test.go
```diff
@@ -262,6 +243,118 @@ var _ = Describe("Out", func() {
...
+		Context("when the registry returns 429 Too Many Requests", func() {
+			var registry *ghttp.Server
+
+			BeforeEach(func() {
+				registry = ghttp.NewServer()
+
+				layers, err := randomImage.Layers()
+				Expect(err).ToNot(HaveOccurred())
+
+				// digest, err := fakeImage.Digest()
```
kcmannem: leftovers?

xtreme-sameer-vohra:
Testing
Dockerfile
```
FROM alpine:edge
#COPY --from=builder assets/ /opt/resource/
RUN dd if=/dev/urandom of=file.txt bs=1048576 count=100
```
Test script
```
#!/bin/bash
for i in {1..500}
do
    echo "Running $i time"
    docker build . --no-cache -t SOME_REPO/test
    docker save -o /tmp/foo.tar SOME_REPO/test
    ./out /tmp/foo.tar < /tmp/req
done
```
req
```
{
	"source": {
		"password": "SOME_PASSS",
		"repository": "SOME_REPO/test",
		"username": "SOME_USER"
	}
}
```


In 155 attempts, we saw 2 errors, of which 1 is worth noting;
## 1
2019/11/26 05:14:27 existing blob: sha256:754950418dd96f778edc10a609207bf829015a276f1c628a4a3bb076f1eeba87
2019/11/26 05:14:30 pushed blob: sha256:f2e23afd6b56affb5a7cef763b9f5ce181a88798300873e09769d3d0b7e93fd5
ERRO[0279] pushing image failed: upload image: Patch https://index.docker.io/v2/xtremesameer/test/blobs/uploads/ea54c6f6-a0c1-4316-8776-8b19e8d8e863?_state=26HykXIdTWsld4UsL01kJfHwtzTGmjqgegAlJOGrKrx7Ik5hbWUiOiJ4dHJlbWVzYW1lZXIvdGVzdCIsIlVVSUQiOiJlYTU0YzZmNi1hMGMxLTQzMTYtODc3Ni04YjE5ZThkOGU4NjMiLCJPZmZzZXQiOjAsIlN0YXJ0ZWRBdCI6IjIwMTktMTEtMjZUMTA6MTQ6MzEuNjIyODMxNzdaIn0%3D: io: read/write on closed pipe

## 2
Successfully built a9b5cec0dab7
Successfully tagged xtremesameer/test:latest
INFO[0001] pushing sha256:603ae342cec979f0017aea951b11bb9e1b0b27aa7c1906ef88a170baf154d89e to index.docker.io/xtremesameer/test:latest
2019/11/25 23:24:11 existing blob: sha256:754950418dd96f778edc10a609207bf829015a276f1c628a4a3bb076f1eeba87
2019/11/25 23:24:14 pushed blob: sha256:a9b5cec0dab7c7d4569c0d0f3c1fcd0bd37ffad831540367c94edf12ba180524
ERRO[0284] pushing image failed: upload image: BLOB_UPLOAD_UNKNOWN: blob upload unknown to registry; map[]

PR #96, opened 2020-02-07
Adds support for ECR auth when running concourse on an EC2 instance

nealedj:
I run concourse on an EC2 instance which has an instance role which already allows it ECR permissions.

The existing implementation would have forced me to some static credentials available within concourse and I'd rather not if possible as credentials are already available from the metadata API.

There's a similar PR open on the s3-resource: https://github.com/concourse/s3-resource/pull/115

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,6 +48,10 @@ differences:
 * `aws_role_arn`: *Optional. Default `"`.* If set, then this role will be
   assumed before authenticating to ECR.
 
+* `aws_use_ec2_role_creds`: *Optional. Default `"`.* If set, then the role of the
+  current EC2 instance will be used to authenticate to ECR. Use this when running
+  concourse on EC2.
+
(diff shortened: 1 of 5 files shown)


hjribeiro:
+1

vito:
Sorry but I can't merge this because it breaks the security model of sharing version history.

More context here: https://github.com/concourse/concourse/issues/3023

tl;dr: Resources are identified entirely by their configuration. If credentials are not in the configuration, the version history will become 'shared' to people who shouldn't have access. Future plans for supporting ECR auth are described in #3023.

Thanks for the PR though. :slightly_smiling_face: You're welcome to use a fork until proper support is implemented, I just don't want to adopt this pattern in our core resource types.

analytically:
This breaks our flow as well. The docker-image breaks in 6.3.0 and I don't want to be adding credentials in my pipelines.

PR #98, opened 2020-02-10
Use ECR without AwsAccessKeyId and AwsSecretAccessKey

### LINKED ISSUE #94: Allow ECR access without aws_access_key_id and aws_secret_access_key
I've been trying out the ECR features that were just added, thanks! Our workers already have an assumed role, so they don't need to auth using an access key. PR coming shortly

william-tran:
This allows ECR to be used if a role is already assumed. Fixes #94

### DIFF
--- a/cmd/check/main.go
+++ b/cmd/check/main.go
@@ -42,7 +42,7 @@ func main() {
 		return
 	}
 
-	if req.Source.AwsAccessKeyId != "" && req.Source.AwsSecretAccessKey != "" && req.Source.AwsRegion != "" {
+	if req.Source.AwsRegion != "" {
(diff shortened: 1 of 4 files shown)


nealedj:
Nice to see somebody else had the same issue as me :)

I also raised a few last week to fix this: #96

I went with using an explicit parameter to get the resource to authenticate to ECR rather than relying on the region being set. Not sure which solution is preferable?

vito:
Sorry, but I can't merge this change - see https://github.com/concourse/registry-image-resource/pull/96#issuecomment-603831222 for an explanation. Thanks for the PR!

Issue #92, opened 2020-02-05
Support `skip_download`?


amnk:
Currently, when using `skip_download`, this resource produces ```ERRO[0000] invalid payload: json: unknown field "skip_download" ```

Will it be possible to add that?

absconcept:
I hope we can resolve this soon

FlorianSW:
As a user of both, the registry-image and the docker-image, resources, and given the following key difference of registry-image to docker-image:
```
A goal of this resource is to stay as focused and simple as possible. The Docker Image resource grew way too large and complicated. There are simply too many ways to build and publish Docker images. It will be easier to support many smaller resources + tasks rather than one huge interface.
```
I would be super interested in the use case of such a flag :D You just said, that it does not exist, which is true, but for a real consideration, if this resource should've such a functionality, I think the actual use case (so, what you want to achieve) would be very very helpful :)

v-yarotsky:
I have a simple pipeline:
- Grab new version of a repo's master branch
- Build the docker image
- Push the docker image 
- Deploy the image to k8s

Downloading the docker image within the pipeline is of no use to me. While uploading takes seconds because of the docker registry layer cache, downloading takes a minute or more sometimes, slowing my team's CI cycle unnecessarily.

I have dozens of these pipelines. Having a flag `skip_download` would speed all of them up considerably, which I know would be greatly appreciated by my teammates.

FlorianSW:
Alright, that makes sense, thanks for providing your use case :)

YoussB:
#181 merged, closing.

PR #103, opened 2020-02-14
Add support of `skip_download` to avoid pulling the image

amnk:
Fixes concourse/registry-image-resource#92

### DIFF
--- a/cmd/in/main.go
+++ b/cmd/in/main.go
@@ -60,37 +60,42 @@ func main() {
 		logrus.SetLevel(logrus.DebugLevel)
 	}
 
-	if len(os.Args) < 2 {
-		logrus.Errorf("destination path not specified")
-		os.Exit(1)
-		return
-	}
+	if req.Params.Download() == true {
 
-	dest := os.Args[1]
-
-	if req.Source.AwsAccessKeyId != "" && req.Source.AwsSecretAccessKey != "" && req.Source.AwsRegion != "" {
-		if !req.Source.AuthenticateToECR() {
+		if len(os.Args) < 2 {
+			logrus.Errorf("destination path not specified")
 			os.Exit(1)
 			return
 		}
-	}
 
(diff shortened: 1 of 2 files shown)


dgrizzanti:
We're looking to take advantage of this feature. What's holding up merging this? Happy to help with changes if needed.

Thanks!

amnk:
I think it just needs a review and approval. Maybe @vito can check it or @YoussB

types.go
```diff
@@ -235,6 +236,14 @@ func (p GetParams) Format() string {
...
 }
 
+func (p GetParams) Download() bool {
+	if !p.SkipDownload {
+		return true
+	}
+
+	return !p.SkipDownload
+}
+
```
YoussB, line 246: I don't think it's necessary to create a method for this - looking at how it gets called, it should be enough to reference the `SkipDownload` struct field directly. It's an understandable instinct to create seam for this functionality in case further business logic gets added later, but let's favour simplicity for the time being.

cmd/in/main.go
```diff
@@ -60,37 +60,42 @@ func main() {
...
+		err = resource.RetryOnRateLimit(func() error {
+			return get(req, ref, dest)
+		})
+		if err != nil {
+			logrus.Errorf("fetching image failed: %s", err)
+			os.Exit(1)
+			return
+		}
+	} else {
+		logrus.Info("Skipping download because `skip_donwload` is set to `true``")
```
YoussB, line 98: could we invert this conditional and instead have an early return, like
```go
       if req.Params.SkipDownload {
               logrus.Info("Skipping download because `skip_download` is set to `true``")
               json.NewEncoder(os.Stdout).Encode(InResponse{
                       Version:  req.Version,
                       Metadata: req.Source.Metadata(),
               })
               return
       }
```
This would result in a smaller diff overall.

YoussB:
Hey @amnk,
I am going to go ahead and close this for now, but will appreciate you reopening when needed.
I will be more more than happy to continue reviewing once you address the above comments.

Thanks!

PR #125, opened 2020-03-13
Fix Go 1.14 support

tustvold:
As per the 1.14 release notes [here](https://golang.google.cn/doc/go1.14)

> Number no longer accepts invalid numbers, to follow the documented behavior more closely.

This is especially problematic as the golang builder image is only major version pinned, so currently the docker images don't build


fenech:
@tustvold sorry, I didn't see this before creating #132 (now merged) so I guess that this one can be closed now.

vito:
Ah yep didn't see this one. Closing!

PR #132, opened 2020-03-22
Fix tag parsing for Go 1.14

fenech:
The behaviour of json.Number was changed in Go 1.14 to no longer
accept invalid
numbers (https://golang.org/doc/go1.14#encoding/json). This means that
it can no longer be used to handle both strings and numbers, as we
were doing here.

Instead, first try parsing as a string, and if that doesn't work,
convert from number to string via RawMessage.


fenech:
Note that the unit tests on master were broken without this, since they use the base image `golang:1` that currently points to 1.14.1.

PR #183, opened 2020-05-20
Added `RegistryMirror` Source attribute

### LINKED ISSUE #37: Feature request: mirror registry
Hello,
would you consider supporting a (clearly optional) mirror parameter?

I know that the old `docker-image-resource` did support that: https://github.com/concourse/docker-image-resource/pull/42/files

If you think that's a valid use case I "might" be able to contribute.
Thanks!

owenfarrell:
Fixes #37

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,6 +54,11 @@ differences:
 * `debug`: *Optional. Default `false`.* If set, progress bars will be disabled
   and debugging output will be printed instead.
 
+* `registry_mirror`: *Optional.*
+  * `host`: *Required.* A hostname pointing to a Docker registry mirror service.
+  * `username` and `password`: *Optional.* A username and password to use when
+  authenticating to the mirror.
+
 * `content_trust`: *Optional.* Configuration about content trust.
   * `server`: *Optional.* URL for the notary server. (equal to `DOCKER_CONTENT_TRUST_SERVER`)
   * `repository_key_id`: *Required.* Target key's ID used to sign the trusted collection, could be retrieved by `notary key list`
--- a/cmd/check/main.go
+++ b/cmd/check/main.go
@@ -57,33 +57,37 @@ func main() {
 	}
 
 	var response CheckResponse
-
-	origin := repo.Tag(req.Source.Tag())
+	tag := new(name.Tag)
 
 	if req.Source.RegistryMirror != nil {
-		registry, err := name.NewRegistry(req.Source.RegistryMirror.Host, name.WeakValidation)
+		origin := repo.Registry
+
+		mirror, err := name.NewRegistry(req.Source.RegistryMirror.Host, name.WeakValidation)
 		if err != nil {
 			logrus.Errorf("could not resolve registry: %s", err)
 			os.Exit(1)
 			return
 		}
 
-		repo.Registry = registry
-		mirror := repo.Tag(req.Source.Tag())
+		repo.Registry = mirror
+		*tag = repo.Tag(req.Source.Tag())
 
-		response, err = checkWithRetry(req.Source.RegistryMirror.BasicCredentials, req.Version, mirror)
+		response, err = checkWithRetry(req.Source.RegistryMirror.BasicCredentials, req.Version, *tag)
 		if err != nil {
 			logrus.Warnf("checking mirror %s failed: %s", mirror.RegistryStr(), err)
 		} else if len(response) == 0 {
 			logrus.Warnf("checking mirror %s failed: tag not found", mirror.RegistryStr())
 		}
+
+		repo.Registry = origin
 	}
 
-	if response == nil || len(response) == 0 {
-		response, err = checkWithRetry(req.Source.BasicCredentials, req.Version, origin)
+	if len(response) == 0 {
+		*tag = repo.Tag(req.Source.Tag())
+		response, err = checkWithRetry(req.Source.BasicCredentials, req.Version, *tag)
 	}
 	if err != nil {
-		logrus.Errorf("checking origin %s failed: %s", origin.RegistryStr(), err)
+		logrus.Errorf("checking origin %s failed: %s", tag.RegistryStr(), err)
 		os.Exit(1)
 		return
 	}
--- a/cmd/check/main.go
+++ b/cmd/check/main.go
@@ -49,21 +49,41 @@ func main() {
 		}
 	}
 
-	ref, err := name.ParseReference(req.Source.Name(), name.WeakValidation)
+	repo, err := name.NewRepository(req.Source.Repository, name.WeakValidation)
 	if err != nil {
-		logrus.Errorf("could not resolve repository/tag reference: %s", err)
+		logrus.Errorf("failed to resolve repository: %s", err)
 		os.Exit(1)
 		return
 	}
 
 	var response CheckResponse
-	err = resource.RetryOnRateLimit(func() error {
-		var err error
-		response, err = check(req, ref)
-		return err
-	})
+
+	origin := repo.Tag(req.Source.Tag())
+
+	if req.Source.RegistryMirror != nil {
+		registry, err := name.NewRegistry(req.Source.RegistryMirror.Host, name.WeakValidation)
+		if err != nil {
+			logrus.Errorf("could not resolve registry: %s", err)
+			os.Exit(1)
+			return
+		}
+
+		repo.Registry = registry
+		mirror := repo.Tag(req.Source.Tag())
+
+		response, err = checkWithRetry(req.Source.RegistryMirror.BasicCredentials, req.Version, mirror)
+		if err != nil {
+			logrus.Warnf("checking mirror %s failed: %s", mirror.RegistryStr(), err)
+		} else if len(response) == 0 {
+			logrus.Warnf("checking mirror %s failed: tag not found", mirror.RegistryStr())
+		}
+	}
+
+	if response == nil || len(response) == 0 {
+		response, err = checkWithRetry(req.Source.BasicCredentials, req.Version, origin)
+	}
 	if err != nil {
-		logrus.Errorf("check failed: %s", err)
+		logrus.Errorf("checking origin %s failed: %s", origin.RegistryStr(), err)
 		os.Exit(1)
 		return
 	}
@@ -71,10 +91,20 @@ func main() {
 	json.NewEncoder(os.Stdout).Encode(response)
 }
 
-func check(req CheckRequest, ref name.Reference) (CheckResponse, error) {
+func checkWithRetry(principal resource.BasicCredentials, version *resource.Version, ref name.Tag) (CheckResponse, error) {
+	var response CheckResponse
+	err := resource.RetryOnRateLimit(func() error {
+		var err error
+		response, err = check(principal, version, ref)
+		return err
+	})
+	return response, err
+}
+
+func check(principal resource.BasicCredentials, version *resource.Version, ref name.Tag) (CheckResponse, error) {
 	auth := &authn.Basic{
-		Username: req.Source.Username,
-		Password: req.Source.Password,
+		Username: principal.Username,
+		Password: principal.Password,
 	}
 
 	imageOpts := []remote.Option{}
@@ -101,11 +131,8 @@ func check(req CheckRequest, ref name.Reference) (CheckResponse, error) {
 	}
 
 	response := CheckResponse{}
-	if req.Version != nil && !missingTag && req.Version.Digest != digest.String() {
-		digestRef, err := name.ParseReference(req.Source.Repository+"@"+req.Version.Digest, name.WeakValidation)
-		if err != nil {
-			return CheckResponse{}, fmt.Errorf("resolve repository/digest reference: %w", err)
-		}
+	if version != nil && !missingTag && version.Digest != digest.String() {
+		digestRef := ref.Repository.Digest(version.Digest)
 
 		digestImage, err := remote.Image(digestRef, imageOpts...)
 		var missingDigest bool
@@ -122,7 +149,7 @@ func check(req CheckRequest, ref name.Reference) (CheckResponse, error) {
 				return CheckResponse{}, fmt.Errorf("get cursor image digest: %w", err)
 			}
 
-			response = append(response, *req.Version)
+			response = append(response, *version)
 		}
 	}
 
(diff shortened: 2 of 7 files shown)


owenfarrell:
A couple of open questions:

The unit tests currently fail because `registry-1.docker.io` isn't a legitimate mirror. Is the `concourse/test-image-static` published anywhere else that is accessible?

xtreme-sameer-vohra:
Hey @owenfarrell 
Thank you for submitting the PR. I am going to take a look and add a review shortly.

xtreme-sameer-vohra:
I think it would be sufficient to test that the mirror registry was queried in the tests. One option would be to use the `go-containerregistry` pkg which has a [registry](https://github.com/google/go-containerregistry/blob/master/cmd/registry/main.go).

Would you be open to updating the tests to leverage it or explore another option ?

xtreme-sameer-vohra:
> I think it would be sufficient to test that the mirror registry was queried in the tests. One option would be to use the `go-containerregistry` pkg which has a [registry](https://github.com/google/go-containerregistry/blob/master/cmd/registry/main.go).
> 
> Would you be open to updating the tests to leverage it or explore another option ?

Actually a simpler approach might be do so something like;
```
req.Source = resource.Source{
					Repository: "fakeserver.foo:1111/concourse/test-image-static",
					RawTag:     "latest",
					RegistryMirror: "docker.io",
				}
```
use a non-existent address for the repository definition and the fully qualified one for the mirror. This ensures, the resource is able to check against the mirror.

Let me know if that makes sense to you.

owenfarrell:
@xtreme-sameer-vohra - Thanks for the feedback. Your approach makes total sense. I was able to refactor the `in` and `check` unit tests and validate them locally. Validating `out` took a bit more work, but I think I got there.

On the latest PR build, there is still [one failure](https://ci.concourse-ci.org/teams/main/pipelines/registry-image/jobs/prs-alpine/builds/171#L5ecbd0f9:209), but it looks like its unrelated to the changes I made (and it passed in the [previous build](https://ci.concourse-ci.org/teams/main/pipelines/registry-image/jobs/prs-alpine/builds/170#L5eca67e6:358) 🤷‍♂️ ).

owenfarrell:
@xtreme-sameer-vohra Thanks for the feedback. I added some new unit tests to cover the fallback scenario that you described, using a fake/real combination for repository and mirror.

The logic I added will retry on _**any**_ error, not just a missing image repository. Keeping it generic/simple that way could cover other scenarios that would be beneficial (mirror is down, repository exists but digest is missing, etc.).

cmd/check/main.go
```diff
@@ -64,19 +65,31 @@ func main() {
...
-	err = resource.RetryOnRateLimit(func() error {
+	op := func() error {
 		var err error
 		response, err = check(req, ref)
 		return err
-	})
+	}
+	err = resource.RetryOnRateLimit(op)
+	if err != nil {
+		logrus.Warnf("checking mirror %s failed: %s", ref.RegistryStr(), err)
```
xtreme-sameer-vohra, line 83: The business logic makes sense, but, this would show an incorrect log messages if a `mirror` isn't configured and the `origin` fails.
Pulling out the `op` function looks good, I am wondering if the function would be easier to read if the `mirror` logic as under a single `if` block ?

cmd/in/main.go
```diff
@@ -99,18 +100,30 @@ func main() {
...
 	fmt.Fprintf(os.Stderr, "fetching %s@%s\n", color.GreenString(req.Source.Repository), color.YellowString(req.Version.Digest))
 
-	err = resource.RetryOnRateLimit(func() error {
+	op := func() error {
 		return get(req, ref, dest)
-	})
+	}
+	err = resource.RetryOnRateLimit(op)
+	if err != nil {
+		logrus.Warnf("fetching mirror %s failed: %s", ref.RegistryStr(), err)
```
xtreme-sameer-vohra, line 117: Same as the comment on `check/main.go`

cmd/check/main.go
owenfarrell: > this would show an incorrect log messages if a mirror isn't configured and the origin fails.

Good call. Definitely needs an update.

> I am wondering if the function would be easier to read if the mirror logic [was] under a single if block ?

The part that I'm struggling with here is how you signal the need for a retry against the origin. In order to move the `RetryOnRateLimit` call up in to the `if` block, we need a way to determine if another call to `RetryOnRateLimit` is needed (and skipping the follow on try if the first was successful).

I'll play around with it a bit, but if you have something in mind and can sketch out an example, that would help me follow what you're thinking.

README.md
```diff
@@ -113,7 +115,8 @@ In this format, the resource will produce the following files:
 
 ### `out`: Push an image up to the registry under the given tags.
 
-Uploads an image to the registry under the tag configured in `source`.
+Uploads an image to the registry (or `registry_mirror`, if defined) under the
```
xtreme-sameer-vohra, line 118: 👍

cmd/check/main.go
```diff
@@ -49,29 +49,53 @@ func main() {
...
+		origin = new(name.Registry)
+		*origin = repo.Registry
+		repo.Registry = registry
+	}
+
+	ref := repo.Tag(req.Source.Tag())
+
 	var response CheckResponse
-	err = resource.RetryOnRateLimit(func() error {
+	op := func() error {
```
xtreme-sameer-vohra, line 76: I missed this earlier, but I am thinking its not desirable for the function to be a closure but rather have its arguments called out in the signature.

owenfarrell: Are you suggesting that the `op` variable be a function with 1 parameter (`ref`)? Or something else?

Since there's a similar strategy used in `in`, I'm assuming the same change would apply there, too.

owenfarrell:
> Expected: it to fetch the image from origin as the private registry doesn't have the image.
Actual: the private registry responds but with an empty result as opposed to an error, which results in the origin not being queried.

I didn't have my Docker network set up correctly in the tests before. I was able to reproduce this by running the tests directly from my machine against a containerized (empty) registry. Should be fixed now.

> * attempt to push against `origin` if `mirror` errors

This comment made me realize one unintended implication of this change: the potential for leaking credentials.

The unit tests for `put` all require a username and password. I think there are few scenarios (if any) where access to a mirror and an origin share the same credentials. Implementing a fallback mechanism that assumes the use of the same principal and the same level of access seems suspect.

For example:
1. an enterprise-hosted mirror of Docker Hub or GCR requires enterprise credentials... but those credentials won't work on Hub or GCR. Blindly sending those credentials to a different host seems unnecessary.
1. ECR uses a completely different authentication mechanism compared to just-about-everything-else.

The same issue holds for `check` and `get` as well. 🤔 

Part of me wonders if we shouldn't revisit one of your first comments:

> out when configured with a mirror will only push to the mirror, it doesn't push the image to the origin as well. I suspect this is okay behaviour, but we can document it in the README.md so folks aren't surprised.


How would you feel about changing course a bit on this one:

1. Expand the `RegistryMirror` field to be a `struct` instead of a string (to support its own credentials)
1. Revert the changes to `put` such that the resource only support pushes to `origin` (and assume that any/all mirrors will replicate as needed)

xtreme-sameer-vohra:
Those are valid points and your recommendations for changing course make sense too 👍 

> How would you feel about changing course a bit on this one:
> 
> 1. Expand the `RegistryMirror` field to be a `struct` instead of a string (to support its own credentials)
> 2. Revert the changes to `put` such that the resource only support pushes to `origin` (and assume that any/all mirrors will replicate as needed)

xtreme-sameer-vohra:
cc @vito in case you have any suggestions/thoughts

vito:
Just catching up, but pushing to the origin instead of the mirror sounds better to me. Pushing to a mirror sounds wonky since you'd expect it to purely reflect the state in the origin and not have any unique data on its own. 🙂

owenfarrell:
@xtreme-sameer-vohra - I took another crack at the implementation.

There was one extra implication of moving away from the closures and parameterizing the private functions for `in`. By splitting the retrieval of the `v1.Image` (metadata) from the retrieval of the blobs creates a de facto transaction checkpoint mid-way through the function.

So there is some impact to the `in_test` since `in` no longer needs to re-fetch the metadata once the metadata fetch is successful.

xtreme-sameer-vohra:
Thanks for the PR @owenfarrell and especially for your patience through the back-and-forth 👏

evanchaoli:
I cloned this repo, made a private repo and tested it. It actually didn't work for me.

I setup a private registry of mirror with the standard `registry` image, and I can pull from this mirror with raw `docker pull` as well as `docker-image` resource, but `registry-image` always hit error:

```
fetching busybox@sha256:2131f09e4044327fd101ca1fd4043e6f3ad921ae7ee901e9142e6e36b354a907
WARN[0001] fetching mirror 10.185.19.14:5000 failed: locate remote image: UNKNOWN: unknown error; map[errors:[map[code:DENIED message:requested access to the resource is denied] map[code:UNAUTHORIZED detail:[map[Action:pull Class: Name:busybox Type:repository]] message:authentication required]]] 
```

I was very confused with the error `authentication required`. Initially, I tried to add `insecure` option:

```golang
mirror, err := name.NewRegistry(req.Source.RegistryMirror.Host, name.WeakValidation, name.Insecure)
```

But still got same error. Thought I spent time reading the code, I couldn't figure out the cause. Then I managed to capture packets from my private registry, and the problem appeared immediately. Look at the HTTP packet this `registry-image` sent to the mirror:

registry-image-miss-namesapce

the path missed namespace of repository. When I used `docker pull`, the path looked like:

```
GET /v2/library/busybox/....
```

Then I added a piece of code in `in.go`:

```golang
const defaultNamespace = "library"

	if req.Source.RegistryMirror != nil && !strings.ContainsRune(req.Source.Repository, '/') {
		req.Source.Repository = fmt.Sprintf("%s/%s", defaultNamespace, req.Source.Repository)
	}
```

Now, it worked with my private mirror.

@owenfarrell @xtreme-sameer-vohra @vito  Do you guys agree with my analysis and my fix?

owenfarrell:
@evanchaoli - Could you share the source configuration for your `registry-image-resource`?

Since the the `insecure` attribute disables TLS, I wouldn't expect that to resolve a `401: Unauthorized` response code.

Based on the error, it sounds like your mirror requires some kind of authentication (e.g. a `username` and `password`).

owenfarrell:
Thanks for the post.

The underlying go-containerregistry library has a [similar block](https://github.com/google/go-containerregistry/blob/master/pkg/name/repository.go#L34-L37) of logic to your proposal. However, it only prepends the `library` namespace when specifically interacting with Docker Hub. The `defaultNamespace` is not prepended to any other registries.

Taking a quick glance through Docker's [V2 API spec](https://docs.docker.com/registry/spec/api/), there's no mention of "library" being an implicit namespace. In fact, some of the examples provided explicitly include `library/` in the repository name.

So I think the "proper" answer (adhering to all the specs) would be that an explicit namespace should always be included when using any mirror or a private registry. But the pragmatist in me recognizes that Docker Hub is de facto default and is mirrored more than any other registry.

Does explicitly adding the `library/` namespace to the repository in your pipeline (without the proposed patch) resolve your issue?

evanchaoli:
@owenfarrell Explicitly adding the `library/` works for me without my patch, but I would consider that as unfair. Note that, I'm not just a pipeline author, but I'm a Concourse cluster maintainer. I would also receive our users escalations. A simple argue could be that, when a docker daemon is configured with a mirror, `docker pull busybox` (without `library/`) still works.

The other argue could be, "mirror" implies mirror of docker hub but a private repo, thus "mirror" should be transparent to end users, obviously requesting users to explicitly adding `library/` would break the identical user experience.

So I strongly think `registry-image` should support the default namespace.

owenfarrell:
I don't agree that defining a mirror should imply that it is a mirror of Docker Hub, as container distribution is becoming increasingly fragmented (ECR, GCR, GitHub packages). And I don't agree that emulating the `docker-image-resource` should be a driving factor for functionality, as there is some functionality (like building a container) that is intentionally not being brought in.

However, I do recognize the value of supporting a de facto standard and the impact on users when you don't.

It looks like @vito already has addressed this as part of #230. By doing a deep clone of the `Source` object, and retrieving the `Repository` string, I think implied namespaces will be supported as part of `check` as part of [this commit](https://github.com/concourse/registry-image-resource/pull/230/commits/8475516061b98a8e6131779c68550dbda07efe89#diff-0b27fd093d74ed6ecbed3be722e1a468R67).

However, without an equivalent change to `in`, I think pipelines will be able to `check` for new images in repositories that don't specify a namespace, but they will fail to pull those images.

cmd/in/main.go
```diff
@@ -120,25 +165,38 @@ func get(req InRequest, ref name.Reference, dest string) error {
 		imageOpts = append(imageOpts, remote.WithAuth(auth))
 	}
 
-	image, err := remote.Image(ref, imageOpts...)
+	image, err := remote.Image(digest, imageOpts...)
 	if err != nil {
-		return fmt.Errorf("locate remote image: %w", err)
+		return nil, fmt.Errorf("locate remote image: %w", err)
+	}
+	if image == empty.Image {
```
vito, line 172: @owenfarrell Do you remember what this is about? Noticed it while refactoring and can't see any tests that cover it. Do mirror registries sometimes return empty images?

vito:
> It looks like @vito already has addressed this as part of #230. By doing a deep clone of the `Source` object, and retrieving the `Repository` string, I think implied namespaces will be supported as part of `check` as part of [this commit](https://github.com/concourse/registry-image-resource/pull/230/commits/8475516061b98a8e6131779c68550dbda07efe89#diff-0b27fd093d74ed6ecbed3be722e1a468R67).

Didn't *intend* to address this but good to know! I should probably backfill an explicit test for this. 🤔

> However, without an equivalent change to `in`, I think pipelines will be able to `check` for new images in repositories that don't specify a namespace, but they will fail to pull those images.

Making a similar change there now!

cmd/in/main.go
owenfarrell: @vito - I think that conditional can go.

I originally added it in support of the retry logic in place to support a 429 response. But when testing against a local (dummy) "mirror", I realized that the manifest fetch and blob fetch are really 2 different transactions with a checkpoint in between.

So to make the [test sequence](https://github.com/concourse/registry-image-resource/pull/183/files/8e96807508d46b29c8265c033a07cf5d6533c783#diff-bad64590066f2bc6f879c51275e6a399R398) realistic, I took those requests out... but I never removed that conditional.

Issue #151, opened 2020-04-17
Semver tags should be used for versions - not digests


vito:
This has been brewing for a while. I hope to have more concrete details soon.

Over time there have been a few indicators that the versioning approach of this resource is a bit problematic, and we should instead find a reasonable way to interpret semver-like tags as versions, either instead of or in addition to digests (for purposes of detecting overwritten tags).

Digests are great for precisely identifying an image and detecting when a tag has changed, but pretending that a single tag represents a versioned history is a bit of a lie, since a tag can only really point to one digest, and doesn't contain information about prior digests that it pointed to.

Tags are often used semver-style, with a single repository having tags for e.g. `1.0.0`, `1.0.1`, `1.2.3` etc. This is in addition to tags like `1` and `1.0` which are "shorthand" allowing folks to refer to a semver "version line" without having to be precise. We'd have to figure out how to reconcile those too.

I'm opening this early so I can go back to it later and provide more concrete pain points around the current digest-based versioning scheme, but if you've run into any please leave them here too!

vito:
Some context here: https://github.com/concourse/registry-image-resource/issues/48#issuecomment-517318523

It seems like this should help for situations where we're *pushing* new version tags. The current model forces you to configure a tag beforehand.

vito:
We've been having a lot of trouble pushing tags properly, so it might be a good idea to have the resource do all the heavy lifting.

For example, let's say you have these tags published, with `=` notation here to show the same digest pushed to multiple tags:

* `6.0.0` = `6.0` = `6` = `latest`
* `5.3.1` = `5.3` = `5`
* `5.3.0`
* `5.2.0` = `5.2`

If I'm pushing version `5.2.1`, the resource could know to also push it as `5.2`.

If I'm pushing version `6.1.0`, the resource should know to also push it as `6.1`, `6` and `latest`.

If I'm pushing version `7.0.0`, the resource should know to also push it as `7.0`, `7`, and `latest`.

If I'm pushing version `6.1.1`, the resource should know to also push it as `6.1` and `6`.

vito:
Another common convention is "variant" images, i.e. `6.1.2-alpine`, `6.1-alpine`, `6-alpine`, `alpine` (instead of `latest-alpine`). This is also sometimes done as a prefix rather than a suffix.

vito:
[...]

PR #197, opened 2020-06-05
feat: Add the ability to monitor a registry with a tag filter

jghiloni:
Somewhat related to the conversation(?) in #151, this PR provides the pipeline author a way to monitor a registry image for new tags, and to use that for the get criterion. This has been tested against public and private repos.


jghiloni:
Let's call this WIP for now -- I realized this doesn't help for anything other than check scripts, I want to be able to handle pinning, as well as adding the tag to the in metadata

vito:
@jghiloni Just FYI, #151 has been on my weekend to-do list for a while so I'm taking a swing at it now! Just thought I'd ping in case you had also started looking into it beyond what's done here so far. Happy to hold off if you are.

jghiloni:
Nope, go ahead! This has had to take a backseat.

vito:
@jghiloni #214 is just about ready if you'd like to take a look! Curious to know whether the semver pattern matches your use case.

jghiloni:
It's exactly our use case, would love to give it a shot when you get a chance.

Issue #203, opened 2020-06-12
Alpine image does not honor /etc/hosts due to missing /etc/nsswitch.conf


tvon:
If for whatever reason you need to set `CONCOURSE_GARDEN_ADDITIONAL_HOST_ENTRY` (or populate `/etc/hosts` some other way) then it will be ignored by alpine images.

This is apparently due to https://github.com/golang/go/issues/22846 and I suppose https://github.com/gliderlabs/docker-alpine/issues/367.

In short, `/etc/nsswitch.conf` is not present in alpine which causes golang to ignore `/etc/hosts`. To remedy this the file `/etc/nsswitch.conf` needs to be created with the content `hosts: files dns` (I believe this tells it to look at `/etc/hosts` first and then DNS, given the default is to fallback to `/etc/hosts` only if DNS does not resolve).

The `ubuntu` variant works as expected.

PR #204, opened 2020-06-12
Add /etc/nsswitch.conf so /etc/hosts is read

tvon:
Addresses https://github.com/concourse/registry-image-resource/issues/203

~I have not yet had a chance to test this.~

I pushed a build to `tvon/registry-image-resource:branch-alpine-nsswitch` and was able to confirm that `/etc/hosts` is honored with this change in a Concourse cluster.

### DIFF
--- a/dockerfiles/alpine/Dockerfile
+++ b/dockerfiles/alpine/Dockerfile
@@ -14,6 +14,10 @@ FROM alpine:edge AS resource
 RUN apk add --no-cache bash tzdata ca-certificates unzip zip gzip tar
 COPY --from=builder assets/ /opt/resource/
 RUN chmod +x /opt/resource/*
+# Ensure /etc/hosts is honored
+# https://github.com/golang/go/issues/22846
+# https://github.com/gliderlabs/docker-alpine/issues/367
+RUN echo "hosts: files dns" > /etc/nsswitch.conf
 
 FROM resource AS tests
 COPY --from=builder /tests /tests


YoussB:
Hey @tvon, 
Thanks for the PR, it looks pretty straight forward. I tried building the image and it is working. Have you been able to test it in a concourse environment?
Also, Do you think there is a way we can add a test in the resource's test suite that makes sure that the resource will work under these conditions?

tvon:
Hi @YoussB, I was able to confirm it works in our Concourse environment.

As far as tests go, I can look into adding a red/green that fails without this change.

YoussB:
Hey @tvon,
were you able to create some tests? I would love to merge this :)

tvon:
I got distracted, I'll poke at it today a bit and see what I can figure out.

YoussB:
cool thanks!

tvon:
I commented out my fix but I can't seem to get a failing test.

YoussB:
Thanks @tvon, the change itself looks straight forward, I think it can be merged, but definitely having a test to back it up will be great. Thanks again.

tvon:
I am going to have to bail out on a test for this, I have [a gist which demonstrates `/etc/hosts` being ignored](https://gist.github.com/tvon/8351aaac45071aa927547c3af6b3099b) but I cannot sort out why the test environment will resolve without `/etc/nsswitch.conf`.

YoussB:
the ubuntu build has already passed but not reflected here.

PR #209, opened 2020-06-16
Modify Dockerfiles to allow caching of go mod

YoussB:
- updates the ubuntu and alpine dockerfiles to allow the caching of the
go mod as intermediate containers, this allows for faster builds.

### DIFF
--- a/dockerfiles/alpine/Dockerfile
+++ b/dockerfiles/alpine/Dockerfile
@@ -1,8 +1,12 @@
 FROM golang:1 as builder
-COPY . /src
 WORKDIR /src
+
+COPY go.mod .
+COPY go.sum .
+RUN go mod download
+
+COPY . .
 ENV CGO_ENABLED 0
-RUN go get -d ./...
 RUN go build -o /assets/in ./cmd/in
 RUN go build -o /assets/out ./cmd/out
 RUN go build -o /assets/check ./cmd/check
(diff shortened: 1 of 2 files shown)


dockerfiles/alpine/Dockerfile
```diff
@@ -1,8 +1,10 @@
 FROM golang:1 as builder
 COPY . /src
 WORKDIR /src
+COPY go.mod .
+COPY go.sum .
+RUN go mod download
```
xtreme-sameer-vohra, line 6:
```suggestion
WORKDIR /src
COPY go.mod .
COPY go.sum .
RUN go mod download
COPY . .
```
I think we might need to move these lines above the `COPY` step so that it is cached desirably. Similar to [Concourse Dockerfile](https://github.com/concourse/concourse/blob/master/Dockerfile#L7-L15)

dockerfiles/ubuntu/Dockerfile
```diff
@@ -1,8 +1,10 @@
 FROM concourse/golang-builder as builder
 COPY . /src
 WORKDIR /src
+COPY go.mod .
+COPY go.sum .
+RUN go mod download
```
xtreme-sameer-vohra, line 6:
```suggestion
WORKDIR /src
COPY go.mod .
COPY go.sum .
RUN go mod download
COPY . .
```

Same as comment above

YoussB:
Thanks @xtreme-sameer-vohra, 
I first applied the suggested changes through the github gui but it failed to signoff the commit so I just amended the original commit.

Issue #45, opened 2019-06-18
Feature request: support digest in 'source'


johha:
Hi,
we'd like to specify a digest instead of a tag in the source section of a registry-image resource. This would make it easier for us to ensure that our jobs are using the correct version of our images because tags can be changed easily and only the digest is really unique.

I know that there is [resource pinning](https://github.com/concourse/concourse/issues/2439) available in Concourse v5 but it's not really the same thing.
E.g. if I provide a wrong digest (no image available for that digest) the pinning causes the job to wait forever until a resource with this digest is available. Instead I would like to have a failing resource with some error message (e.g. digest not found).

As a digest is an alternate way to get a specific version of an image I think the right place to specify it would be in source section:
```
resources:
- name: my-image
  type: registry-image
  source:
    repository: myrepo/my-image
    digest: sha256:a252b563fad36e16b71c0aab7d1ea8e380c6a4b21d69d6d38efa4cd873e8da09
```

What do you think about this?

vito:
Hmm it seems like we should instead fix/improve the UX of version pinning in Concourse. I don't want every resource author to have to implement pinning themselves. Would you mind opening a feature request for that instead? Thanks!

jchesterpivotal:
In the example:

```
resources:
- name: my-image
  type: registry-image
  source:
    repository: myrepo/my-image
    digest: sha256:a252b563fad36e16b71c0aab7d1ea8e380c6a4b21d69d6d38efa4cd873e8da09
```

Does `myrepo/my-image@sha256:a252b563fad36e16b71c0aab7d1ea8e380c6a4b21d69d6d38efa4cd873e8da09` not work?

vito:
@jchesterpivotal That won't work since the resource doesn't respect a `digest:` field under `source:`, but this should work:

```yaml
resources:
- name: my-image
  type: registry-image
  source:
    repository: myrepo/my-image
  version:
    digest: sha256:a252b563fad36e16b71c0aab7d1ea8e380c6a4b21d69d6d38efa4cd873e8da09
```

...as well as this (in a task config):

```yaml
image_resource:
  type: registry-image
  source:
    repository: myrepo/my-image
  version:
    digest: sha256:a252b563fad36e16b71c0aab7d1ea8e380c6a4b21d69d6d38efa4cd873e8da09
```

PR #213, opened 2020-06-21
Document how to specify the digest when used as anonymous resource

tlwr:
Implements #45 

What
---

As a user,
when using an anonymous resource in a task,
I want to specify the digest of the image_resource,
so that task runs are reproducible:

```
platform: linux
image_resource:
  type: docker-image
  source:
    repository: golang
    digest: 'sha256:b55ababe05036f7d28086d00f5ea585baf4373a20d81979e0f202ced998443a8'
run:
  path: date
```

This is especially useful when the maintainer of the docker image:

- does not publish tags
- re-uses tags

Why
---

I understand the reasoning in #45, namely that pinning is better. I agree that specifying a digest is bad practice for most resources, as they are supposed to be dynamic (unless pinned).

However I think the registry-image-resource (and docker-image-resource) is a special exception as [anonymous resources](https://concourse-ci.org/tasks.html#schema.anonymous_resource) used in tasks. As anonymous resources are checked when the task runs, a user does not have the opportunity to pin them, or control their versioning other than via tags.

### DIFF
--- a/README.md
+++ b/README.md
@@ -174,6 +174,29 @@ The currently encouraged way to build these images is by using the
   list of tag values to tag the image with (in addition to the tag configured
   in `source`).
 
+### Use in tasks
+
+Images used as
+[image resources](https://concourse-ci.org/tasks.html#schema.task.image_resource)
+in tasks are called
+[anonymous resources](https://concourse-ci.org/tasks.html#schema.anonymous_resource).
+Anonymous resources can specify
+[a version](https://concourse-ci.org/tasks.html#schema.anonymous_resource.version),
+which is the image digest. For example:
+
+
+```
+image_resource:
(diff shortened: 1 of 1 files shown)


tlwr:
@taylorsilva thank you for your detailed answer

I've dropped all the code-changes, and have updated the README with a section for how to specify the digest.

tlwr:
Sorry for necro-ing this PR

I need to update this to fix the typo `docker-image` should be `registry-image`

Secondly, I found myself trying to use this today for a resource type rather than a resource, is this possible - [it doesn't appear to be](https://github.com/concourse/concourse/blob/master/atc/config.go#L188-L199) - maybe at that point it is time to host your own image 😓  (which is what I ended up doing)

PR #214, opened 2020-06-22
add semver tagging semantics

### LINKED ISSUE #151: Semver tags should be used for versions - not digests

vito:
fixes #151 
closes #197 (supersedes it)

still a WIP - leaving lots of context in each commit message, will amend this with more commentary later

note: this is intended to be as backwards-compatible as possible, but the default behavior will change specifically when `tag:` is not specified. rather than tracking only the `latest` tag, the resource will track all semver tags present, only returning `latest` if there are no semver tags with the same digest

TODO:

* [x] `/check`: support detecting "variants" (i.e. `alpine`, `3-alpine`, `3.2-alpine`, `3.2.4-alpine`)
* [x] `/check`: distinguish between multi-word variants (i.e. `apache` shouldn't match `1.2.3-foo-apache`)
* [x] `/check`: distinguish between prereleases and variants (i.e. `apache` SHOULD match `1.2.3-rc.2-apache`)
* [x] `/check`: skip prereleases by default, add flag
* [x] `/check`: support checking "from" a version
* [x] `/check`: refactor; there's a lot of branching with tag vs. non-tag and mirror vs. no mirror
* [x] `/out`: support pushing versions and variant versions
* [x] (#230): `/out`: implement optional semver tag bumping (i.e. bump `3.2`, `3`, and `latest` when shipping `3.2.4` assuming `3.2` is latest)
  * [x] (#230) plus variant support (bump `3.2.0-alpine`, and optionally `3.2-alpine` if latest 3.2.x, `3-alpine` if latest 3.x, `alpine` if latest overall)
* [x] update README
* [x] An initial `check` request which fetches digests of all the tags is a bit slow - it looks like `go-containerregistry` does a fresh auth flow for each call. Can we optimize this, either by avoiding requests when given a cursor version or by opening an issue on `go-containerregistry`? (opened issue: https://github.com/google/go-containerregistry/issues/740)
* [ ] ~~`/in`: respect a version that only has `digest:` - this is may be from a user manually pinning, per https://github.com/concourse/registry-image-resource/issues/45#issuecomment-647573611~~
  * hmm, actually maybe we should just always `check` and then find the specified version?


vito:
@xtreme-sameer-vohra Quick update: this is good to go! Just finished the last TODO item and fixed a broken test.

README.md
```diff
@@ -1,6 +1,20 @@
...
 Supports checking, fetching, and pushing of images to Docker registries.
+This resource can be used in two ways: [with `tag`
+specified](#check-with-tag-discover-new-digests-for-the-tag) and [without
+`tag`](#check-without-tag-discover-semver-tags).
+
+With `tag` specified, `check` will detect changes to the digest the tag points
+to, and `out` will always push to the specified tag. This is to be used in
+simpler cases where no real versioning exists.
+
+With `tag` omitted, `check` will instead detect tags based on semver versions
```
xtreme-sameer-vohra, line 12: Since semver will be the default, will existing resources in pipelines that don't specify a `tag` will continue to maintain the current status quo where `latest` is the default ?

vito: Yep. It works a bit differently, since it'll change to detecting all the semver versions, and if `latest` has the same digest as a semver tag (which is also the latest version) you'll see that tag instead, but `latest` will always take precedence if it has its own special digest or if there are no semver tags at all.

So, this should be backwards compatible, with one edge case exception: if `latest` points to the same digest as a tag `1.2.3` but there is a newer version tag, say `1.3.0`, the resource will return `1.3.0` whereas before it would have returned the `1.2.3` digest.

This is a little awkward to fix because it's a state that the registry could be in temporarily while a new version is being bumped - say `1.3.0` has just shipped but the `latest` tag hasn't bumped yet. If we have `latest` take priority, the version history would go from (1.2.3) to (1.2.3, 1.3.0, latest [1.2.3]), and then finally to (1.2.3, 1.3.0). Which seems a bit janky.

vito:
One thing that might be worth looking into before merging: when used in `image_resource` and we run a `check`, if the `check` is not given a version this will return all versions instead of just the latest one. This may become expensive if run too often. Can we make it so `check` checks from the last version? 🤔

vito:
Following up on my previous comment, it does look like this could introduce extra request load on the registry, since [we only check from the *given* version, if present](https://github.com/concourse/concourse/blob/a790d822ccd757c59ba7ec1dbbf336e4bc6e6c14/atc/worker/image/image_resource_fetcher.go#L332).

cmd/check/main.go
```diff
@@ -49,66 +46,259 @@ func main() {
...
+}
+
+type TagVersions []TagVersion
+
+func (vs TagVersions) Len() int           { return len(vs) }
+func (vs TagVersions) Less(i, j int) bool { return vs[i].Version.LessThan(vs[j].Version) }
+func (vs TagVersions) Swap(i, j int)      { vs[i], vs[j] = vs[j], vs[i] }
+
+func checkTag(source resource.Source, version *resource.Version) (resource.CheckResponse, error) {
+	repo, err := name.NewRepository(source.Repository, name.WeakValidation)
```
xtreme-sameer-vohra, line 292: This can be DRY'ed up with checkRepository

cmd/check/main.go
```diff
@@ -49,66 +46,259 @@ func main() {
...
+	}
+
+	auth := &authn.Basic{
+		Username: source.Username,
+		Password: source.Password,
+	}
+
+	imageOpts := []remote.Option{}
+
+	rt, err := transport.New(repo.Registry, auth, http.DefaultTransport, []string{repo.Scope(transport.PullScope)})
```
xtreme-sameer-vohra, line 125: is this required for `remote.List`  and therefore isn't used in checkTag ?

README.md
xtreme-sameer-vohra: That makes sense. The probability of the repo being in the state of the edge case is low and based on the meaning of `latest`, the returning  `1.3.0` as the last version rather than `1.2.3` makes more sense logically.

cmd/check/main.go
vito: It isn't as important for `checkTag` because we'll only make one digest request (or two if the digest changed) in that checking mode. This is only here because the semver tag checking involves one digest request per semver tag. This commit provides some explanation: https://github.com/concourse/registry-image-resource/pull/214/commits/611afb71cf7091f6656d1a10859c1a7e79691b13

I could consolidate them anyway if not just to DRY up the code as you suggested. The difference would be that instead of making an extra auth token request we'd make an extra ping request, which is still better.

cmd/check/main.go
vito: done!

xtreme-sameer-vohra:
> Following up on my previous comment, it does look like this could introduce extra request load on the registry, since [we only check from the _given_ version, if present](https://github.com/concourse/concourse/blob/a790d822ccd757c59ba7ec1dbbf336e4bc6e6c14/atc/worker/image/image_resource_fetcher.go#L332).

Yep, its also expensive in terms of time - took ~17-19 against concourse/concourse

cmd/check/main.go
```diff
@@ -49,108 +46,284 @@ func main() {
...
+		if from != nil && identifier == from.Tag && digest.String() == from.Digest {
+			// if the 'from' version exists and has the same digest, treat its
+			// version as a cursor in the tags, only considering newer versions
+			//
+			// note: the 'from' version will always be the first one hit by this loop
+			cursorVer = ver
+		}
+	}
+
+	sort.Sort(semver.Collection(versions))
```
xtreme-sameer-vohra, line 243: This doesn't appear to be reference subsequently and `versions` can be dropped.

cmd/in/main.go
```diff
@@ -92,54 +72,59 @@ func main() {
 		return
 	}
 
-	fmt.Fprintf(os.Stderr, "fetching %s@%s\n", color.GreenString(req.Source.Repository), color.YellowString(req.Version.Digest))
+	tag := repo.Tag(req.Version.Tag)
 
-	var image v1.Image
-	digest := new(name.Digest)
+	if !req.Params.SkipDownload {
```
xtreme-sameer-vohra, line 77: Hey, I might have missed the purpose of refactoring in/main.go, moving the request and response struct definitions looks good. But it appears as though the the main block has one extra level of nesting to be able to move the response writing section to the bottom.

IMHO, might be simpler and more readable to extract the response writing to a reusable private function and keep the main block flatter. WDYT ?

cmd/out/main.go
```diff
@@ -70,32 +61,74 @@ func main() {
...
+		}
+
+		tagsToPush = append(tagsToPush, repo.Tag(tag))
+	}
+
 	for _, tag := range tags {
 		n := fmt.Sprintf("%s:%s", req.Source.Repository, tag)
 
-		extraRef, err := name.ParseReference(n, name.WeakValidation)
+		extraRef, err := name.NewTag(n, name.WeakValidation)
```
xtreme-sameer-vohra, line 118:
```suggestion
		extraTag, err := name.NewTag(n, name.WeakValidation)
```

cmd/out/main.go
```diff
@@ -70,32 +61,74 @@ func main() {
...
+
+	repo, err := name.NewRepository(req.Source.Repository)
 	if err != nil {
-		logrus.Errorf("could not parse additional tags: %s", err)
+		logrus.Errorf("could not resolve repository: %s", err)
 		os.Exit(1)
 		return
 	}
 
-	var extraRefs []name.Reference
```
xtreme-sameer-vohra, line 87: noice, keeping it simpler with Tags rather than Refs

cmd/out/main.go
```diff
@@ -130,65 +163,56 @@ func main() {
...
-		Metadata: req.Source.MetadataWithAdditionalTags(tags),
+		Metadata: append(req.Source.Metadata(), resource.MetadataField{
+			Name:  "tags",
+			Value: strings.Join(pushedTags, " "),
+		}),
 	})
 }
 
-func put(req OutRequest, img v1.Image, ref name.Reference, extraRefs []name.Reference) error {
+func put(req resource.OutRequest, img v1.Image, refs []name.Tag) error {
```
xtreme-sameer-vohra, line 192:
```suggestion
func put(req resource.OutRequest, img v1.Image, tags []name.Tag) error {
```

cmd/out/main.go
```diff
@@ -130,65 +163,56 @@ func main() {
...
-
-		err = trustedRepo.SignImage(img)
-		if err != nil {
-			logrus.Errorf("failed to sign image: %s", err)
-		}
 	}
 
-	for _, extraRef := range extraRefs {
-		logrus.Infof("pushing as tag %s", extraRef.Identifier())
+	for _, extraRef := range refs {
```
xtreme-sameer-vohra, line 207:
```suggestion
	for _, tag := range tags {
```

cmd/out/main.go
```diff
@@ -130,65 +163,56 @@ func main() {
...
+		logrus.Infof("pushing to tag %s", extraRef.Identifier())
 
-		err = remote.Write(extraRef, img, remote.WithAuth(auth), remote.WithTransport(http.DefaultTransport))
+		err = remote.Write(extraRef, img, remote.WithAuth(auth))
```
xtreme-sameer-vohra, line 210:
```suggestion
		err = remote.Write(tag, img, remote.WithAuth(auth))
```

cmd/out/main.go
```diff
@@ -130,65 +163,56 @@ func main() {
...
-		err = trustedRepo.SignImage(img)
-		if err != nil {
-			logrus.Errorf("failed to sign image: %s", err)
-		}
 	}
 
-	for _, extraRef := range extraRefs {
-		logrus.Infof("pushing as tag %s", extraRef.Identifier())
+	for _, extraRef := range refs {
+		logrus.Infof("pushing to tag %s", extraRef.Identifier())
```
xtreme-sameer-vohra, line 208:
```suggestion
		logrus.Infof("pushing to tag %s", tag.Identifier())
```

cmd/in/main.go
vito: Ah right - sorry, at one point I was trying to fix #227 but it looks like I committed some partial work here.

If you don't mind the additional content to review, I think I'll carry forward with that so I can refactor the `RegistryMirror` usage in `in` to match the refactor I made to `check` at the same time.

vito:
@xtreme-sameer-vohra Converting this back to a draft for now. Thanks for the reviews so far!

TODO:

* [ ] Integrate suggested changes from above
* [ ] Remove `WeakValidation` in places where it's already the default
* [ ] `check` should only return the latest version after all. 🙁 (Unless we change image resource checking to check from the last known version.)
* [ ] Explicit test for handling of `library` with mirrors, per discussion started here: https://github.com/concourse/registry-image-resource/pull/183#issuecomment-662833877
* [ ] Figure out what's up with the `empty.Image` handling per https://github.com/concourse/registry-image-resource/pull/183#discussion_r460266597

owenfarrell:
> Explicit test for handling of library with mirrors, per discussion started here: #183 (comment)

I put #231 out there as a way to address this issue, but I targeted #230 (not `master`). But now I'm not sure which PR is "winning", this one or #230.

Happy to rebase those changes over whichever PR is the go-forward approach.

vito:
@owenfarrell Thanks! It would be better to target this PR. #230 is supposed to be based on it; I split it out so there'd be less to review. But I've since pushed more changes to this PR, so it's out of sync.

Alternatively you could target `master` - I'm happy to integrate the same fix here.

vito:
@xtreme-sameer-vohra I merged `master` into this instead of rebasing. There are a bunch of commits here so rebasing would have been a nightmarish hellscape vs. merging and resolving all the conflicts at once. Hope that's OK.

vito:
@xtreme-sameer-vohra This is ready to go again. We're thinking of jumping to v7.0 instead of v6.8 so I'd like to include this too since we're doing a major version bump and this involves a behaviour change, even though it's *mostly* backwards-compatible.

I wish I could make it easier to review but with how long this has been open it'd be a bit of a nightmare to rebase. All of my recent pushes are just to bring it in line with changes that have been made on `master` - mainly the registry mirror handling and the `HEAD`/`GET` fallback, and previously the whole `commands/` restructuring. So hopefully there's not a whole lot to review from last time. 🙏

cmd/check/main.go
```diff
@@ -166,16 +171,38 @@ func checkRepository(principal resource.BasicCredentials, version *resource.Vers
...
+			pre := ver.Prerelease()
+			if pre != "" {
+				// contains additional variant
+				if strings.Contains(pre, "-") {
+					continue
+				}
+
+				if !strings.HasPrefix(pre, "alpha.") &&
+					!strings.HasPrefix(pre, "beta.") &&
+					!strings.HasPrefix(pre, "rc.") {
```
aoldershaw: What about tags that don't follow this format exactly? e.g. https://hub.docker.com/r/kubernetesui/dashboard/tags?page=1&name=v2.0.0-rc7, doesn't have a `.` after `rc`? This may not be a valid semver tag, not sure - but it looks like `1.2.3-alpha` is, but wouldn't be supported

aoldershaw:
Just skimmed through so far, but seems pretty nice!

Have you considered adding a [semver constraint](https://github.com/Masterminds/semver#checking-version-constraints) configuration to the `source`? Seems like it'd be super easy with Masterminds/semver

cmd/check/main.go
vito: good point - i think that's easy to support. pushed a fix!

README.md
```diff
@@ -109,15 +129,108 @@ registry_key: |
...
+
+#### Pre-release versions
+
+By default, pre-release versions are ignored. With `pre_releases: true`, they
+will be included.
+
+Note however that variants and pre-releases both use the same syntax:
+`1.2.3-alpine` is technically also valid syntax for a Semver prerelease. For
+this reason, the resource will only consider prelease data containing
+`alpha.`, `beta.`, or `rc.` as a proper prerelease, treating anything else as
```
aoldershaw, line 227:
```suggestion
this reason, the resource will only consider prerelease data starting with
`alpha`, `beta`, or `rc` as a proper prerelease, treating anything else as
```

vito:
> Have you considered adding a [semver constraint](https://github.com/Masterminds/semver#checking-version-constraints) configuration to the `source`? Seems like it'd be super easy with Masterminds/semver

Yeah I think that makes a lot of sense. I'd rather leave it for a separate PR though, just 'cause this one's pretty big already.

commands/check.go
```diff
@@ -107,30 +86,202 @@ func (c *check) Execute() error {
...
+	Version *semver.Version
+}
+
+type TagVersions []TagVersion
+
+func (vs TagVersions) Len() int           { return len(vs) }
+func (vs TagVersions) Less(i, j int) bool { return vs[i].Version.LessThan(vs[j].Version) }
+func (vs TagVersions) Swap(i, j int)      { vs[i], vs[j] = vs[j], vs[i] }
+
+func checkTag(tag name.Tag, source resource.Source, version *resource.Version, opts ...remote.Option) (resource.CheckResponse, error) {
```
aoldershaw, line 272: GitHub won't let me comment on the line I want to, but:

I think we need to put the `tag` in the version here too, since now `in` populates `./tag` based on the version

aoldershaw: The following pipeline demonstrates this:

```yaml
resource_types:
- name: registry-image-pr
  type: registry-image
  source:
    repository: concourse/registry-image-resource
    tag: pr-alpine-214

resources:
- name: nginx
  type: registry-image-pr
  source:
    repository: nginx
    tag: alpine

jobs:
- name: job
  plan:
  - get: nginx
  - task: task
    config:
      platform: linux
      image_resource:
        type: registry-image
        source: {repository: busybox}
      inputs:
      - name: nginx
      run:
        path: cat
        args: [./nginx/tag]
```

it also looks weird in the UI, as it has an empty tag in metadata

README.md
```diff
@@ -26,12 +40,18 @@ differences:
...
+
+  *Note: If using ecr you only need the repository name, not the full URI e.g.
+  `alpine` not `012345678910.dkr.ecr.us-east-1.amazonaws.com/alpine`*
+
+* `tag`: *Optional.* Instead of monitoring semver tags, monitor a single tag
+  for changes (based on digest).
 
-* `tag`: *Optional. Default `latest`.* The name of the tag to monitor and
-  publish to.
+* `variant`: *Optional.* Detect only tags matching this variant suffix, and
```
aoldershaw, line 52: It might be obvious, but *maybe* it's worth pointing out that this doesn't do anything if you have `tag` set, and that you should instead add the variant to the `tag`.

e.g. you'd specify:

```
tag: stable-alpine
```

rather than

```
tag: stable
variant: alpine
```

vito: It technically still does something - it won't be used for `check`, but if someone does a `put` with a `version` set, the configured variant will be appended to the version:

```yaml
resources:
- name: my-image
  type: registry-image
  source:
    repository: vito/foobar
    tag: banana
    variant: alpine

jobs:
- name: pusha-t
  plan:
  - put: my-image
    params: {version: 1.2.3} # pushes 'banana' and '1.2.3-alpine'
```

It's a little weird, but it's a thing.

Pushed a README change, lemme know what you think!

commands/check.go
vito: fixed!

README.md
aoldershaw: Ah thanks, makes sense

aoldershaw:
`check` is certainly doing a lot more work, given that it has to make a `HEAD` request for each (semver) tag. I wanted to see how much slower it is:

```
$ jq -n '{source:{repository:"nginx"}}' | \
    time docker run --rm -i concourse/registry-image-resource:pr-alpine-214 \
      /opt/resource/check
...
       25.66 real         0.05 user         0.02 sys
```

compared to latest:

```
$ jq -n '{source:{repository:"nginx"}}' | \
    time docker run --rm -i concourse/registry-image-resource \
      /opt/resource/check
...
        1.53 real         0.05 user         0.02 sys
```

It's quite a bit slower. I expect this won't be a huge problem for named resources, since they only need to do this slow check once (and whenever a digest disappears) - but will it be for anonymous `task.image_resource`s?

vito:
@aoldershaw Yup - that's what led to the whole image-fetching refactor. With how things used to work, it would definitely be too slow for task image resources - but now that `check` runs from the latest version, it should only be slow the first time the image is ever used.

Scopes are a little confusing, but in this case you can ignore them since scopes only apply to pipeline-level resources - otherwise it'll be the "null" (global) scope.

aoldershaw (approved):
LGTM!

PR #230, opened 2020-07-10
bumping aliases

vito:
this feature makes it trivial to bump 'alias' tags such as `latest`, `alpine`, `1`, `1.2` when pushing `1.2.3` or `1.2.3-alpine` - see README changes for details!

### DIFF
--- a/commands/out.go
+++ b/commands/out.go
@@ -107,81 +107,12 @@ func (o *Out) Execute() error {
 		tagsToPush = append(tagsToPush, repo.Tag(tag))
 
 		if req.Params.BumpAliases && ver.Prerelease() == "" {
-			auth := &authn.Basic{
-				Username: req.Source.Username,
-				Password: req.Source.Password,
-			}
-
-			imageOpts := []remote.Option{}
-
-			if auth.Username != "" && auth.Password != "" {
-				imageOpts = append(imageOpts, remote.WithAuth(auth))
-			}
-
-			versions, err := remote.List(repo, imageOpts...)
+			aliasTags, err := aliasesToBump(req, repo, ver)
 			if err != nil {
-				return fmt.Errorf("list repository tags: %w", err)
+				return fmt.Errorf("determine aliases: %w", err)
 			}
 
-			bumpLatest := true
-			bumpMajor := true
-			bumpMinor := true
-			for _, v := range versions {
-				versionStr := v
-				if req.Source.Variant != "" {
-					versionStr = strings.TrimSuffix(versionStr, "-"+req.Source.Variant)
-				}
-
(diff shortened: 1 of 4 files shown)


vito:
@chenbh #214 has been merged, so this is ready for review!

Once this is done I can (finally) get back to https://github.com/concourse/concourse/issues/5792

out_test.go
```diff
@@ -405,13 +405,183 @@ var _ = DescribeTable("pushing semver tags",
...
+		// rationale: tags are representing versions of a cohesive product, so
+		// its versions should be comparable across variants.
+		SemverTagPushExample{
+			Tags: []string{"1.3.0"},
+
+			Variant:     "foo",
+			Version:     "1.2.3",
+			BumpAliases: true,
+
+			PushedTags: []string{"1.2.3-foo", "1.2-foo"},
```
chenbh, line 484: Just for sanity
```suggestion
			PushedTags: []string{"1.2.3-foo", "1.2-foo", "foo"},
```

chenbh (changes requested):
Nit: can you refactor the versioning logic into a separate function? A 200+ line function is kinda off-putting to me

chenbh:
Also took a *very* surface look at content trust, seems like it's fine since we only use it to sign when pushing, not verify image/tags when checking/pulling

commands/out.go
```diff
@@ -105,6 +105,84 @@ func (o *Out) Execute() error {
...
+				Password: req.Source.Password,
+			}
+
+			imageOpts := []remote.Option{}
+
+			if auth.Username != "" && auth.Password != "" {
+				imageOpts = append(imageOpts, remote.WithAuth(auth))
+			}
+
+			versions, err := remote.List(repo, imageOpts...)
```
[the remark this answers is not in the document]
vito: I don't think we want to account for registry mirrors; `put` doesn't use the mirror because it needs to make sure it's talking to the source of truth.

out_test.go
vito: Not sure what you mean here - this would be counter to what the test is saying. We *don't* want `foo` bumped because 1.2.3 is not the latest version of the overall product; 1.3.0 is. Do you think this should behave differently?

vito:
Extracted that big chunk of code out. 👍

out_test.go
vito: Oo, good point - I hadn't thought of a situation where you'd want some variants to be intentionally older. I was only thinking of the `alpine`/`ubuntu` case where they represent different 'flavors' of the same product.

I'll change it to the proposed behavior. 👍

chenbh (approved):
lgtm!

PR #231, opened 2020-07-24
Add support for implicit namespaces when pulling from mirrors

owenfarrell:
@vito - So I wasn't right after all, you didn't _**quite**_ cover the scenario.

I added tests to cover some scenarios that are already working (miss on mirror, exists in origin). But added a couple of mocked up registry test cases to validate the URL paths submitted for each.

cc: @evanchaoli

### DIFF
--- a/check_test.go
+++ b/check_test.go
@@ -115,21 +115,42 @@ var _ = Describe("Check", func() {
 				})
 
 				Context("which is missing the image", func() {
-					BeforeEach(func() {
-						req.Source = resource.Source{
-							Repository: "concourse/test-image-static",
-							Tag:        "latest",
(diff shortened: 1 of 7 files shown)


evanchaoli:
@owenfarrell Thanks for the fix. I downloaded the patch and tested it locally. It worked perfectly. This fix will be super helpful.

Also, Your solution of getting repo for mirror from origin `name.NewRepository(repo.String())` is much better than hard-code default namespace.

xtreme-sameer-vohra:
Hey @owenfarrell 
Thanks for submitting the PR. I'll review it in the by EOD tomorrow.

owenfarrell:
closing in favor of #232

PR #237, opened 2020-09-08
Call out that aws_region is required for ECR

mattdodge:
It wasn't clear to me that you had to include an `aws_region` if you wanted this resource to know to talk to an AWS ECR repository. Maybe this was dumb user error on my part but without an example of how this would work with ECR I figure this might help others too.

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,7 +46,7 @@ differences:
 * `aws_session_token`: *Optional. Default `""`.* The session token to use
   for authenticating with STS credentials with ECR.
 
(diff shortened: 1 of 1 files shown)


xtremerui (approved):

xtremerui:
@mattdodge LGTM. Could you signed your commit to fix the DCO error? Thx.

mattdodge:
@xtremerui yep no problem. I made my change with the GitHub markdown editor, I thought that signed the commits but I guess not. Just force pushed the signed commit

PR #241, opened 2020-09-17
check: use HEAD request to avoid rate limit

vito:
fixes https://github.com/concourse/concourse/issues/6075


xtreme-sameer-vohra (approved):
lgtm !!

xtremerui:
@vito i think the issue mentioned here https://github.com/concourse/docker-image-resource/issues/307 also applied to registry image resource.

vito:
@xtremerui Yep, thanks for the heads-up. I think it's probably safest to have it only use `HEAD` if it detects that it's talking to Docker Hub or something. Doesn't seem like a trustworthy API for other implementations.

Issue #247, opened 2020-10-15
check attempts to first check the 'registry_mirror', even when a registry-host-prefixed image is specified in the 'repository'


mdb:
In current implementation, if a `registry_mirror` is configured, images whose `source.repository` contains a registry-host-prefixed image (such as `my-registry.com/foo/bar`) will still be first checked at the `registry_mirror` registry and _not_ the registry specified in the `repository` value. This cause 2 unwanted side effects...

* It generates excess requests against the `registry_mirror` in cases where the mirror does not mirror the registry whose host name is specified in the `repository` value, but instead mirrors a _different_ registry.
* I suspect it could fetch _the wrong_ image if the `registry_mirror` happens to host a namespace/image combination that collides with one hosted in the registry specified by the host-prefixed `repository`, but which is in fact a _different_ image altogether

This is especially problematic when a Concourse-wide [default registry has been configured](https://github.com/concourse/concourse/pull/6079). 

For example, given the following default `registry-image` configuration:

```yaml
registry-image:
  registry_mirror:
    host: my-default-mirror.com
```

...and the following `registry-image-resource`:

```yaml
- name: my-image
  type: registry-image
  source:
    repository: my-registry.com/foo/bar
```

...the resource attempts to _first_ check the image at `my-default-mirror.com/foo/bar`. Unless `my-default-mirror.com` is a mirror to `my-registry.com`, the `check` invocation may result in either excess requests to `my-default-mirror.com` or even an incorrect image resolution, as described above.

It's perhaps worth noting that this is not true of `docker` itself, which appears to successfully resolve `my-registry.com/foo/bar` from `my-registry.com`, even when `my-default-mirror.com` has been configured as a registry mirror. In the case of `docker`, it appears `my-registry.com/foo/bar` is successfully resolved without performing any requests to `my-default-mirror.com`.

This is similar to `docker-image-resource` [issue 305](https://github.com/concourse/docker-image-resource/issues/305)

vito:
merged #248 - thanks!

PR #248, opened 2020-10-15
registry_mirror no longer overwrites a repo host

mdb:
Fixes issue #247.

This ensures that, even if a `registry_mirror` is
configured, images whose `source.repository` contains
an explicitly declared registry-host-prefixed image
(such as `my-registry.com/foo/bar`) will be checked at
and fetched from the explicitly declared registry,
and not at the `source.registry_mirror`.

Signed-off-by: Mike Ball <mikedball@gmail.com>

### DIFF
--- a/check_test.go
+++ b/check_test.go
@@ -98,11 +98,38 @@ var _ = Describe("Check", func() {
 				mirror.Close()
 			})
 
+			Context("when the repository contains a registry host name prefixed image", func() {
+				BeforeEach(func() {
+					mirror.AppendHandlers(
+						ghttp.CombineHandlers(
+							ghttp.VerifyRequest("GET", "/v2/"),
+							ghttp.RespondWith(http.StatusOK, `welcome to zombocom`),
+						),
+						ghttp.CombineHandlers(
+							ghttp.VerifyRequest("HEAD", "/v2/some/fake-image/manifests/latest"),
+							ghttp.RespondWith(http.StatusOK, `{"fake":"manifest"}`, http.Header{
+								"Docker-Content-Digest": {LATEST_FAKE_DIGEST},
+							}),
+						),
+					)
+
+					req.Source.Repository = mirror.Addr() + "/some/fake-image"
+					req.Source.RegistryMirror = &resource.RegistryMirror{
+						Host: "thisregistrymirrordoesnotexist.nothing",
+					}
+				})
+
+				It("it checks and returns the current digest using the registry declared in the repository and not using the mirror", func() {
+					Expect(res).To(Equal([]resource.Version{
+						{Digest: LATEST_FAKE_DIGEST},
+					}))
+				})
+			})
+
 			Context("which has the image", func() {
 				Context("in an explicit namespace", func() {
 					BeforeEach(func() {
-						// use the mock mirror as the "origin", use Docker Hub as a "mirror"
-						req.Source.Repository = mirror.Addr() + "/" + req.Source.Repository
+						// use Docker Hub as a "mirror"
 						req.Source.RegistryMirror = &resource.RegistryMirror{
 							Host: name.DefaultRegistry,
 						}
@@ -257,8 +284,7 @@ var _ = Describe("Check", func() {
 			Context("which has the image", func() {
 				Context("in an explicit namespace", func() {
 					BeforeEach(func() {
-						// use the mock mirror as the "origin", use Docker Hub as a "mirror"
-						req.Source.Repository = mirror.Addr() + "/" + req.Source.Repository
+						// use Docker Hub as a "mirror"
 						req.Source.RegistryMirror = &resource.RegistryMirror{
 							Host: name.DefaultRegistry,
 						}
@@ -425,8 +451,7 @@ var _ = Describe("Check", func() {
 			Context("which has the image", func() {
 				Context("in an explicit namespace", func() {
 					BeforeEach(func() {
-						// use the mock mirror as the "origin", use Docker Hub as a "mirror"
-						req.Source.Repository = mirror.Addr() + "/" + req.Source.Repository
+						// use Docker Hub as a "mirror"
 						req.Source.RegistryMirror = &resource.RegistryMirror{
 							Host: name.DefaultRegistry,
 						}
@@ -595,8 +620,7 @@ var _ = Describe("Check", func() {
 			Context("which has the image", func() {
 				Context("in an explicit namespace", func() {
 					BeforeEach(func() {
-						// use the mock mirror as the "origin", use Docker Hub as a "mirror"
-						req.Source.Repository = mirror.Addr() + "/" + req.Source.Repository
+						// use Docker Hub as a "mirror"
 						req.Source.RegistryMirror = &resource.RegistryMirror{
 							Host: name.DefaultRegistry,
 						}
--- a/README.md
+++ b/README.md
@@ -55,7 +55,7 @@ differences:
   and debugging output will be printed instead.
 
 * `registry_mirror`: *Optional.*
-  * `host`: *Required.* A hostname pointing to a Docker registry mirror service.
+  * `host`: *Required.* A hostname pointing to a Docker registry mirror service. Note that this is only used if no registry hostname prefix is specified in the `repository`. If the `repository` contains a registry hostname prefix -- such as `my-registry.com/foo/bar` -- the `registry_mirror` is ignored and the explicitly declared registry in the `repository` value is used.
   * `username` and `password`: *Optional.* A username and password to use when
   authenticating to the mirror.
 
--- a/commands/check.go
+++ b/commands/check.go
@@ -68,7 +68,9 @@ func (c *check) Execute() error {
 	var response CheckResponse
 	tag := new(name.Tag)
 
-	if req.Source.RegistryMirror != nil {
+	// only use the RegistryMirror as the Registry if the repo doesn't use a different,
+	// explicitly-declared, non-default registry, such as 'some-registry.com/foo/bar'.
+	if req.Source.RegistryMirror != nil && repo.Registry.String() == name.DefaultRegistry {
 		mirror, err := name.NewRepository(repo.String())
 		if err != nil {
 			return fmt.Errorf("could not resolve mirror repository: %s", err)
--- a/commands/in.go
+++ b/commands/in.go
@@ -3,6 +3,12 @@ package commands
 import (
 	"encoding/json"
 	"fmt"
+	"io"
+	"io/ioutil"
+	"log"
+	"os"
+	"path/filepath"
+
 	resource "github.com/concourse/registry-image-resource"
 	"github.com/fatih/color"
 	"github.com/google/go-containerregistry/pkg/authn"
@@ -13,11 +19,6 @@ import (
 	"github.com/google/go-containerregistry/pkg/v1/remote"
 	"github.com/google/go-containerregistry/pkg/v1/tarball"
 	"github.com/sirupsen/logrus"
-	"io"
-	"io/ioutil"
-	"log"
-	"os"
-	"path/filepath"
 )
 
 type InRequest struct {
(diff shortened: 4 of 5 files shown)


mdb:
@vito I'm curious to hear your thoughts on this and whether, if approved, a new `registry-image-resource:latest` might be released within a comfortable window ahead of the [November 1, 2020 rate limits](https://www.docker.com/blog/scaling-docker-to-serve-millions-more-developers-network-egress/) Docker Hub plans to impose.

I ask because I believe the ability to configure a default `registry-image-resource.registry_mirror.host` will help mitigate the impending [November 1, 2020 rate limiting](https://www.docker.com/blog/scaling-docker-to-serve-millions-more-developers-network-egress/). However, I'm concerned that doing so without the changes proposed in this PR will have undesired consequence for those users of the `registry-image-resource` whose configurations include a `repository` containing explicitly declared registry-host-prefixed image (such as `my-registry.com/foo/bar`).

vito:
@mdb Yeah, we're aiming to get another release out before that deadline. Thanks for the PR! I'll review this very soon.

vito:
Interesting. It seems like the original code makes sense assuming the mirror is configured individually, but it does indeed become an issue when defaults are applied to the type itself and the type is then used for many different resources configuring different registries. Nice catch, and thanks for putting together the issues + PRs to fix it!

I think my only concern is it seems like you should still be able to set up a mirror for a registry other than Docker Hub. But maybe that's not common? I really don't know.

It's almost like you'd want to configure a mapping from registry URL to mirror URL. This would make the resource config feel kind of awkward though, because the resource definition will correspond to one registry yet have configuration that implies many are involved, purely because of this global defaulting functionality. 🤔 Maybe it could be added as a separate field which clearly documents that it is intended for configuring defaults?

mdb:
> I think my only concern is it seems like you should still be able to set up a mirror for a registry other than Docker Hub. But maybe that's not common? I really don't know.

I'm not sure either, though I've been _assuming_ it's possible & is done.

> It's almost like you'd want to configure a mapping from registry URL to mirror URL.

That occurred to me as well, though I don't _think_ I see the pursuit of such a thing as something that should block this PR's implementation (though perhaps I'm failing to think it through?). 

Regarding this PR's implementation, and factoring into consideration your above (☝️ ) two comments, I think I continue to believe that, if a `repository` contains a registry hostname (e.g. it is a fully qualified image URL), then the `registry-image-resource` should use the `repository`'s registry hostname and ignore a `registry_mirror.host` (as is currently implemented in this PR). Does that seem reasonable or do your comments above indicate more fundamental concerns with the current implementation?

vito:
@mdb The latter - ignoring the nuances of cluster-wide defaults for a moment, it seems to me that if you configure `registry_mirror.host` it should be respected and interpreted as the mirror for whichever registry is configured in `repository:`. So that's why I'm not so sure about the changes in this PR - it makes sense in the context of cluster-wide defaults, but it might break user expectations in isolation.

I could imagine a scenario for example where a large organization has an internal registry that many (many) teams use, but the teams are expected to set up their own local mirror so that the central registry is not under so much load, or maybe just to speed things up (imagining a team in China accessing a registry hosted in the US). In such a scenario I would expect to configure a registry mirror while pointing `repository:` at the internal registry. Does that make sense?

mdb:
> it seems to me that if you configure `registry_mirror.host` it should be respected and interpreted as the mirror for whichever registry is configured in `repository:`.

☝️ @vito Regarding this, for whatever it's worth, it appears this is _not_ the way `docker` works locally on my laptop. It appears that a registry host name contained in the repository takes precedence _over_ the mirror. For example, I configure `docker` to use a mirror (`my-mirror.com`) which mirrors Docker Hub like so:

```json
cat ~/.docker/daemon.json
{
  "debug" : true,
  "experimental" : false,
  "registry-mirrors": ["https://my-mirror.com"]
}
```

...then, when I `docker pull my-registry.com/foo/bar` (where `my-registry.com` is a registry _other_ than `my-mirror.com`), `docker` resolves the image directly from `my-registry.com` and _does not_ attempt to resolve the image through `my-mirror.com`. I see no requests performed against `my-mirror.com`. 

However, `docker` _does_ attempt to resolve the image through `my-mirror.com` in the following situations where I _do_ see requests issued to `my-mirror.com`.

* `docker pull foo/bar`
* `docker pull my-mirror.com/foo/bar`

Does that change your thinking at all?

> I could imagine a scenario for example where a large organization has an internal registry that many (many) teams use, but the teams are expected to set up their own local mirror so that the central registry is not under so much load, or maybe just to speed things up (imagining a team in China accessing a registry hosted in the US). In such a scenario I would expect to configure a registry mirror while pointing `repository:` at the internal registry. Does that make sense?

Regarding ☝️ this, it does make sense, yes, but those users could do the following and still use their team mirror:

```yaml
- name: foo-bar-image
  type: registry-image
  source:
    # drop the registry host name prefix from the 'repository':
    repository: foo/bar
    registry_mirror:
      host: team-mirror.com
```

And, furthermore, given the current PR implementation, the following would _still_ resolve the `foo/bar` image (assuming it's hosted in `the-registry.com` registry), even though it wouldn't do so through the mirror (in other words, it wouldn't _actively break_ user-facing functionality; it would just generate more load on `the-registry.com` than might be intended):

```yaml
- name: foo-bar-image
  type: registry-image
  source:
    repository: the-registry.com/foo/bar
    registry_mirror:
      host: team-mirror.com
```

😄 And, I know you mention ignoring "the nuances of cluster-wide defaults for a moment," but I'm having a tough time doing so because...

As it stands, I _think_ that, without the changes in this PR, the current implementation of `registry-image-resource` may actively _break_ user-facing functionality (with no clear workaround), IIUC, when a default `registry-image-resource.registry_mirror` is configured (as is currently the cornerstone of my own team's November 1 Docker Hub rate limiting mitigation plans given our own rather large Concourse installation which services thousands of users). For example, consider the following:

```yaml
resource_types:

- name: my-custom-resource
  type: registry-image
  source:
    repository: my-registry.com/foo/bar-resource

resources:

- name: my-custom-resource-instance
  type: my-custom-resource
```

...I believe configuration like this (☝️ ) will result in the following user error without the changes implemented in this PR, as the `check` will fail, as the `registry-image-resource` will attempt to resolve `foo/bar-resource` from the mirror & not its explicitly-configured `repository`:

```
resource type 'my-custom-resource' has no version
```

Because...

```
fly check-resource-type --resource-type my-pipeline/my-custom-resource
checking my-pipeline/my-custom-resource in build 837
error: bad response from event source: 404 Not Found
```

vito:
@mdb Ah ha, yeah that does change my thinking. Thanks for breaking this down for me! It seems like a good rule of thumb to be consistent with the `docker` CLI behavior.

I think where I've been confused is that registry mirrors are only ever set up to mirror a single upstream registry. For example, there is only one `proxy` field and it only points to one registry in the [pull-through cache docs](https://docs.docker.com/registry/recipes/mirror/). It doesn't make sense to mirror multiple registries because then you could very easily end up with name collisions.

From that perspective, it clearly doesn't make sense to use a global default mirror registry for all registries, and it also seems sensible for it to only apply when a registry is not specified, since it's consistent with the Docker CLI behavior and solves the problem of knowing when to use the mirror.

I think the only thing that strikes me as strange is this example:

```yml
- name: foo-bar-image
  type: registry-image
  source:
    # drop the registry host name prefix from the 'repository':
    repository: foo/bar
    registry_mirror:
      host: team-mirror.com
```

...because I would personally assume that `foo/bar` refers to a Docker Hub repository, but in fact it's entirely dependent on wherever your mirror is proxying to. It just seems strange to use the conventional public registry syntax to refer to a private repository, but maybe this is normal and expected and just kind of the way things are with Docker and I'm just not used to it. 🤷‍♂️

Another thing that's odd is that it'll silently ignore a configured mirror. Assuming someone *does* explicitly configure this:

```yml
- name: foo-bar-image
  type: registry-image
  source:
    # drop the registry host name prefix from the 'repository':
    repository: example.com/foo/bar
    registry_mirror:
      host: team-mirror.com
```

...it would be nice if the user could notice that this is effectively a mis-configuration. But I get that it needs to not error so that it can be used as a global default, and putting warnings everywhere would be really noisy. 🤔 I wonder if this should be a separate `default_registry_mirror` field for use when configuring it cluster-wide so that the code can differentiate between the scenarios. What do you think?

mdb:
> ...it would be nice if the user could notice that this is effectively a mis-configuration. But I get that it needs to not error so that it can be used as a global default, and putting warnings everywhere would be really noisy.

Yes; I agree this isn't ideal. I did update the `README`, but perhaps future versions could error -- or warn -- if both `registry_mirror` and a fully qualified registry-host-name-prefixed `repository` are configured? I do recommend _against_ error-ing in such a scenario for now, though, as my big picture goal is: configure the ability to support a default registry mirror (to mitigate the November 1st Docker Hub rate limits) _without_ requiring any action amongst my team's thousands of Concourse users :) 

> I wonder if this should be a separate `default_registry_mirror` field for use when configuring it cluster-wide so that the code can differentiate between the scenarios.

Perhaps, but wouldn't doing so deviate from being "consistent with the docker CLI behavior" which you describe as a "good rule of thumb?" In other words, it's my belief that this PR -- in current implementation -- brings `registry-image-resource`'s behavior inline with that of the `docker` CLI, as I point out [above](https://github.com/concourse/registry-image-resource/pull/248#issuecomment-712436510):

> a registry host name contained in the repository takes precedence over the mirror

vito:
@mdb Sorry, to clarify, I'm totally OK with having both `registry_mirror` and `default_registry_mirror` behave as you describe, with the sole difference that `registry_mirror` would consider it an error if it's configured alongside a `repository:` that has an explicit registry, and `default_registry_mirror` won't.

mdb:
@vito Oh, I see. Thanks for clarifying 😄 . I'm not _too_ opinionated about that, though I do wonder if your `default_registry_mirror` idea is...

* Perhaps unnecessary extra code/logic/overhead given that the current PR 248 implementation works just as `docker` does (Although, sidebar: please fact check me on all my assertions about `docker`'s behavior 😄 ) 
* Confusing for users to squint at the not-quite-obvious-and-kinda-hard-to-grok difference between `default_registry_mirror` and `registry_mirror`

By contrast, I think it's a bit easier in code, as well as a bit easier on users' brains to [say in the README](https://github.com/concourse/registry-image-resource/pull/248/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R58)...

```
Note that this is only used if no registry hostname prefix is specified in the `repository`. If the `repository` contains a registry hostname prefix -- such as `my-registry.com/foo/bar` -- the `registry_mirror` is ignored and the explicitly declared registry in the `repository` value is used.
```

...but I can go with the flow if you disagree :)

check_test.go
```diff
@@ -98,11 +98,38 @@ var _ = Describe("Check", func() {
...
+					}))
+				})
+			})
+
 			Context("which has the image", func() {
 				Context("in an explicit namespace", func() {
 					BeforeEach(func() {
-						// use the mock mirror as the "origin", use Docker Hub as a "mirror"
-						req.Source.Repository = mirror.Addr() + "/" + req.Source.Repository
+						// use Docker Hub as a "mirror"
```
vito, line 132: I think we can probably just delete some of these `in an explicit namespace` contexts since the above test kind of makes them redundant, but I can just do that in a followup commit since I've got work in progress in this part of the code anyway. 👍

vito (approved):
No worries! That works for me - it's consistent with the `docker` CLI anyhow (just checked), so maybe it'll be intuitive enough as it is. I'll go ahead and merge this - thanks for the PR!

check_test.go
mdb: Thanks @vito - I appreciate the patience on all this. And I have a naive question: is `registry-image-resource:latest` continuously published to Docker Hub on merge to `master`? Or do I need to wait for the next Concourse release before this is published to Docker Hub? Or some other release event?

vito: Likewise!

Images are built from `master` and pushed to `concourse/registry-image-resource:dev-ubuntu` and `concourse/registry-image-resource:dev-alpine` by [this pipeline](https://ci.concourse-ci.org/teams/main/pipelines/registry-image).

I could also publish a new version but I think I'll wait until I've made this same fix here: https://github.com/concourse/docker-image-resource/pull/308 - so I can ship it all together. In any case there'll be a new version shipped and included in v6.7.0.

PR #249, opened 2020-10-19
check: try GET when HEAD request returns no digest

vito:
some registries don't return Docker-Content-Digest, so in that scenario we'll need to fetch the actual manifest

context: https://github.com/concourse/docker-image-resource/pull/307, https://github.com/concourse/docker-image-resource/pull/308

### DIFF
--- a/check_test.go
+++ b/check_test.go
@@ -87,6 +87,51 @@ var _ = Describe("Check", func() {
 			})
 		})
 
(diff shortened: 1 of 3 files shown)


YoussB (approved):
LGTM! The code looks good, and I tested the image against the configuration in the issue and it is working 🚀

PR #252, opened 2020-10-26
types: Check if ECR registry has been explicitly specified

### LINKED ISSUE #251: Cross-account ECR access is broken
This issue only affects users doing cross-account things with AWS ECR.
The resource currently prepends the ECR `proxyEndpoint` to `source.repository` after authenticating to ECR.
If the user has already explicitly specified an ECR registry, for example `12345678.dkr.ecr.eu-west-1.amazonaws.com/myimage` the resource fails to return any versions because of the prepended endpoint.

Instead, we should avoid prepending the endpoint if the user has already specified an ECR registry in `source.repository`.

vixus0:
Fixes #251

### DIFF
--- a/types.go
+++ b/types.go
@@ -8,6 +8,7 @@ import (
 	"net/url"
 	"os"
 	"path/filepath"
+	"regexp"
 	"strings"
 
 	"github.com/aws/aws-sdk-go/aws"
@@ -210,7 +211,18 @@ func (source *Source) AuthenticateToECR() bool {
 
 	// Update username and repository
 	source.Username = "AWS"
-	source.Repository = strings.Join([]string{strings.TrimPrefix(*result.AuthorizationData[0].ProxyEndpoint, "https://"), source.Repository}, "/")
+
+	// Only prepend registry URL to repository if it hasn't been explicitly specified
+	matched, err := regexp.MatchString(`\d+\.dkr\.ecr\..+\.amazonaws\.com/.*`, source.Repository)
+
+	if err != nil {
+		logrus.Errorf("regexp error: %s", err)
+		return false
+	}
+
+	if !matched {
+		source.Repository = strings.Join([]string{strings.TrimPrefix(*result.AuthorizationData[0].ProxyEndpoint, "https://"), source.Repository}, "/")
+	}
 
 	return true
 }


YoussB (commented):
Hey @vixus0,

Thanks for the PR.

it kind of feels like this goes against the [original intent](https://github.com/concourse/registry-image-resource/blob/master/README.md#source-configuration) of using the repository as only the name of the image.

the original commit here, though a little old, was done by @mattburgess and @dan-hill2802 so maybe they can provide more context around why it is not expected to have the full url.

vixus0:
Yes I did note that comment which is why I let the resource fall back to using the implicit ECR endpoint when one isn't specified. 

However I feel there's some justification for this change as it's entirely valid to provide a fully qualified image name including the registry URL when pulling from a registry that isn't Docker Hub. I don't think we should break expected behaviour for the sake of convenience. The convenience is untouched in any case 😛 .

We pull images cross-account so that there's one source of truth for them and I don't think that's an unusual use-case.

xtremerui:
@vixus0 is #253 solving your issue? Do you still need this PR or we could close it? Thx.

vixus0:
@xtremerui I still need to try out the new version of registry-image-resource. Is there an image being pushed somewhere that I can use?

xtremerui:
@vixus0 yes there are tag [dev](https://hub.docker.com/layers/concourse/registry-image-resource/dev/images/sha256-28dcf9334c5ce5246b8b049c27accccf66daaaf3cfc8486de02f88121355b37c?context=explore) for alpine and tag [dev-ubuntu](https://hub.docker.com/layers/concourse/registry-image-resource/dev-ubuntu/images/sha256-156133778c283dd538e551f5be4b3182d9579b55b99a8d1c0e15423bd3d1fe00?context=explore) for ubuntu, both of which has the change from #253

vixus0:
@xtremerui hey, so I played around with this a bit and there's only one big issue I have. Using the change I made in this PR, I could use the same variable in my Concourse pipeline to specify the repository for this resource and for the image in my Kubernetes configuration. With the change in #253 I don't think this is possible. One fix I can think of is to dump the full image repository into a file like we do for the digest and tag.

xtremerui:
It seems you might want to sync with master and update your PR? I'd prefer a test in this case so we could know better about the change. Thx!

vixus0:
@xtremerui I'll probably close this PR for the time being and raise a new one for writing the full image name to file alongside the digest. Thanks for the feedback :)

PR #253, opened 2020-10-30
feat: adds support for cross account ecr repositories

cludden:
In order to access images from an ECR repository owned by an account that differs from that of the credentials used, the caller must specify a list of 1 or more AWS account ID's in the `GetAuthorizationToken` call. See docs for more info: https://docs.aws.amazon.com/sdk-for-go/api/service/ecr/#GetAuthorizationTokenInput

### DIFF
--- a/types.go
+++ b/types.go
@@ -15,6 +15,7 @@ import (
 	"github.com/aws/aws-sdk-go/aws/credentials/stscreds"
 	"github.com/aws/aws-sdk-go/aws/session"
 	"github.com/aws/aws-sdk-go/service/ecr"
+	"github.com/aws/aws-sdk-go/service/ecr/ecriface"
 	"github.com/sirupsen/logrus"
 )
 
@@ -25,6 +26,7 @@ type AwsCredentials struct {
 	AwsSecretAccessKey string `json:"aws_secret_access_key,omitempty"`
 	AwsSessionToken    string `json:"aws_session_token,omitempty"`
 	AwsRegion          string `json:"aws_region,omitempty"`
+	AWSECRRegistryId   string `json:"aws_ecr_registry_id,omitempty"`
 	AwsRoleArn         string `json:"aws_role_arn,omitempty"`
 }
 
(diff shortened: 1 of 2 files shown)


xtremerui:
@cludden could you give some context of this PR? Thx.

cludden:
added some additional context in the PR comment

xtremerui:
Might be an unrelated question, how does the auth work when accessing an image from a ECR that owned by other account since you can only provide one set of credentials?

chenbh:
This sounds very similar to #252, the main difference being #252 allows for explicit repo url whereas this one infers the repo url from the auth token

@vixus0 would passing in an auth token in order to access a cross-account ecr repo work for you?

vixus0:
@chenbh Yeah, if I understand correctly this PR allows me to explicitly specify the AWS account ID to pull the image from - that works!

cludden:
Good catch! I didn't realize this resource is building the ECR registry using the first result from the `GetAuthorizationToken` output's `AuthorizationData` list. I've updated the PR to accept a single `aws_ecr_registry_id` string parameter, that if specified, overrides the default behavior.

cludden:
> Might be an unrelated question, how does the auth work when accessing an image from a ECR that owned by other account since you can only provide one set of credentials?
By default, if no registry ids (aws account ids) are specified, the `GetAuthorizationToken` call requests an authorization token for the account to which the requesting credentials belong. This PR modifies the resource to allow the resource to specify an account other than the principal's account id. Additionally, the ECR repository in the other account requires a resource policy granting cross account access. more info here: https://docs.aws.amazon.com/AmazonECR/latest/userguide/security_iam_service-with-iam.html#security_iam_service-with-iam-resource-based-policies

types.go
```diff
@@ -184,6 +185,10 @@ func (source *Source) AuthenticateToECR() bool {
 	client := ecr.New(mySession, &config)
 
 	input := &ecr.GetAuthorizationTokenInput{}
+	if source.AWSECRRegistryId == "" {
```
xtremerui, line 188:
```suggestion
	if source.AWSECRRegistryId != "" {
```

xtremerui (changes requested):
wondering is it possible to add a test to cover this change?

cludden:
slight refactor to add some basic tests around the GetAuthorizationTokenInput modification, left the existing `ECR integration is experimental and untested"` warning in place.

cludden:
bumping for visibility

xtremerui (approved):

PR #254, opened 2020-11-11
RFC: Use default key chain for authentication to the registry

aholyoake-bc:
This should allow simple specialisations of the resource to allow for specific container registry variants without changing the existing behaviour.

For example, to use this with GCR the existing registry-image image can be used as a base layer and then we can copy in a ~/.docker/config.json and optionally a credential manager to authenticate.

Our use case:

We exclusively use google container registry and currently authenticate using the `_json_key:service_account.json` hack described here https://github.com/concourse/docker-image-resource/issues/73#issuecomment-255165396.
A big drawback of this approach is that a developer can read this secret by printing it out in a task.
We want to be reasonably sure that any images that are stored in our production container registry have been built by Concourse so that we have some sort of an audit trail for what has been used to build the image. One approach to achieve this is to use the VM service account / application default credentials to authenticate. Ordinarily you would achieve this by using a credential manager and a docker configuration file.

For the `oci-build-task` and `docker-registry` resources we have a partial hack in place that does just that. i.e. a Dockerfile containing

```
ARG VERSION=latest
FROM vito/oci-build-task:$VERSION

COPY docker-credential-gcr /usr/local/bin/
COPY config.json /root/.docker/config.json
```
and a config file containing:
```
{
  "credHelpers": {
    "eu.gcr.io": "gcr"
  }
}
```

However, the `check` stage of the docker-registry resource does not work with this as it doesn't use docker directly. We'd like to get away from the docker-registry resource either way given that development is frozen on it and it is deprecated.

This PR is largely untested, but I would like some feedback on the approach before continuing.

### DIFF
--- a/commands/check.go
+++ b/commands/check.go
@@ -117,6 +117,8 @@ func performCheck(principal resource.BasicCredentials, version *resource.Version
 
 	if auth.Username != "" && auth.Password != "" {
 		imageOpts = append(imageOpts, remote.WithAuth(auth))
+	} else {
+		imageOpts = append(imageOpts, remote.WithAuthFromKeychain(authn.DefaultKeychain))
 	}
 
 	digest, found, err := headOrGet(ref, imageOpts...)
--- a/commands/in.go
+++ b/commands/in.go
@@ -181,6 +181,8 @@ func get(principal resource.BasicCredentials, digest name.Digest) (v1.Image, err
 
 	if auth.Username != "" && auth.Password != "" {
 		imageOpts = append(imageOpts, remote.WithAuth(auth))
+	} else {
+		imageOpts = append(imageOpts, remote.WithAuthFromKeychain(authn.DefaultKeychain))
 	}
 
 	image, err := remote.Image(digest, imageOpts...)
--- a/commands/out.go
+++ b/commands/out.go
@@ -3,6 +3,10 @@ package commands
 import (
 	"encoding/json"
 	"fmt"
+	"io"
+	"net/http"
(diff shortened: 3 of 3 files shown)


vito:
Thanks for bringing this up - but I don't think we should go forward with this approach, for the same reason we've been rejecting PRs to enable EC2 IAM role based auth for AWS services. https://github.com/concourse/concourse/issues/3023 has a lot of info on this.

The tl;dr is that Concourse expects `source:` to contain credentials, because it would like to treat resources as content-addressable based on their type + source configuration so that equivalent resources across teams/pipelines may be treated as a single resource. This way versions can be shared, greatly reducing the `check` workload on large clusters which re-use resources between pipelines and teams. If the `source:` omits credentials and "magically" detects versions anyway, those versions would be shared to other teams which should not have access. (Granted, those teams would still not be able to actually fetch the bits, but in principle the versions themselves may leak sensitive information.)

We're actually working on concourse/concourse#5229 right now which will allow credentials to be fetched on a worker via extensible var source types and then fed back into `source:` explicitly using traditional `((vars))` syntax in a pipeline. This also has the advantage of keeping resource types simple, since they only have to implement static credential configuration. Once this feature is done it should be possible to implement a var source prototype that acquires GCR credentials by running on the appropriate worker.

Hope that makes sense - happy to answer any questions or dissect flaws in this scheme. For now you may want to just run a fork, since it'll be a little while until this is all done. 🙂

aholyoake-bc:
OK, thanks, we'll do that!

Issue #240, opened 2020-09-16
Need to be able to specify a CA cert like docker-image resource


techgnosis:
Concourse is awesome, thanks guys!

I can't migrate to registry-image because it doesn't let me specify a CA cert like docker-image. Specifying a CA cert is crucial for people who have on-prem registries.

docker-image looks like this..
```
- name: concourse-helper
  type: docker-image
  source:
    repository: ((harbor-domain))/library/concourse-helper
    tag: 1
    ca_certs:
    - domain: ((harbor-domain))
      cert: |
        ((ca_cert))
```

crackcomm:
I found myself patching image and using my own also.

```Patch
diff --git a/dockerfiles/alpine/Dockerfile b/dockerfiles/alpine/Dockerfile
index e6e5bcd..77fdab3 100644
--- a/dockerfiles/alpine/Dockerfile
+++ b/dockerfiles/alpine/Dockerfile
@@ -23,6 +23,9 @@ RUN chmod +x /opt/resource/*
 # https://github.com/gliderlabs/docker-alpine/issues/367
 RUN echo "hosts: files dns" > /etc/nsswitch.conf
 
+COPY ca.pem /usr/local/share/ca-certificates/local-ca.crt
+RUN update-ca-certificates
+
 FROM resource AS tests
 COPY --from=builder /tests /tests
 ADD . /docker-image-resource
```


```yaml
resource_types:
- name: registry-image
  type: docker-image
  source:
    repository: crackcomm/registry-image-resource
```

matthewcosgrove:
Another workaround is to leverage the fact that the containers have their certificates propagated from the Concourse worker host.

This works nicely in a bosh deployment by adding the ca cert to trusted_certs 
```
---
- type: replace
  path: /instance_groups/name=bosh/properties/director/trusted_certs?
  value: |
    ((trusted_ca_cert))
```

For more info see #10 and https://github.com/concourse/concourse/issues/1027. Also https://github.com/concourse/concourse/issues/1938

matthewcosgrove:
Since 6.7.0 there is the ability to [Allow configuring source "defaults" for resource types](https://github.com/concourse/concourse/pull/6079) so you also have the option of using resource defaults https://concourse-ci.org/concourse-web.html#resource-defaults. Also see the "defaults:" part of the schema in https://concourse-ci.org/resource-types.html#schema.resource_type

chriseaton:
This just got me as well. YaARG!

PR #264, opened 2021-02-09
Add ability to specify a CA certificate

Infra-Red:
Allow the resource to accept additional trusted certificates by appending our self-signed certs to a copy of the host trusted CA certs.

Resolves https://github.com/concourse/registry-image-resource/issues/240.

### DIFF
--- a/check_test.go
+++ b/check_test.go
@@ -3,6 +3,7 @@ package resource_test
 import (
 	"bytes"
 	"encoding/json"
+	"encoding/pem"
 	"fmt"
 	"net/http"
 	"os/exec"
@@ -20,6 +21,8 @@ import (
 )
 
 var _ = Describe("Check", func() {
+	var actualErr error
+
 	var req struct {
 		Source  resource.Source
 		Version *resource.Version
@@ -47,11 +50,11 @@ var _ = Describe("Check", func() {
 		cmd.Stdout = outBuf
 		cmd.Stderr = GinkgoWriter
 
-		err = cmd.Run()
-		Expect(err).ToNot(HaveOccurred())
-
-		err = json.Unmarshal(outBuf.Bytes(), &res)
-		Expect(err).ToNot(HaveOccurred())
+		actualErr = cmd.Run()
+		if actualErr == nil {
+			err = json.Unmarshal(outBuf.Bytes(), &res)
+			Expect(err).ToNot(HaveOccurred())
+		}
 	}
 
 	Describe("tracking a single tag", func() {
@@ -68,6 +71,8 @@ var _ = Describe("Check", func() {
 			})
 
 			It("returns the current digest", func() {
+				Expect(actualErr).ToNot(HaveOccurred())
+
 				Expect(res).To(Equal([]resource.Version{
 					{Tag: "latest", Digest: LATEST_STATIC_DIGEST},
 				}))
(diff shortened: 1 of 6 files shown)


vito:
Woops, sorry we didn't see this! For some reason our PR overview tool didn't pick it up. 🤔 Leaving this comment to kick it again - we'll assign it to someone on Monday if no one reviews it before then! Thanks for your work, this is a big missing piece.

vito:
Ah - `PayloadTooLargeError: request entity too large` handling the webhook, probably because of that giant code block. That's a fun edge case. I removed it from the description since we should be able to see the same output in the PR builds anyhow. Hope you don't mind. 👍

commands/out.go
```diff
@@ -245,11 +256,47 @@ func createAuth(req resource.OutRequest) *authn.Basic {
 	}
 }
 
+func createTransport(req resource.OutRequest) (http.RoundTripper, error) {
```
vito, line 259: Maybe we can extract this into a method on `Source`? It looks to be duplicated at the moment with `types.go`.

Actually not sure why the code doesn't just use the `(Source).AuthOptions` method here too, but I haven't looked too closely. 🤔

vito (changes requested):
This is a good start - thanks for also updating the README! It'll need some tests before we can call it done though.

I think we could run a fake registry over HTTPS and show that basic resource `in`, `out`, and `check` calls fail without the cert and succeed with the cert.

We already run a fake registry for the `out` tests so maybe you could look to that for an example. For running a TLS server, you can use [`gomega.NewTLSServer`](https://pkg.go.dev/github.com/onsi/gomega@v1.10.5/ghttp#NewTLSServer) and then access the certificate using [(*httptest.Server).Certificate](https://golang.org/pkg/net/http/httptest/#Server.Certificate) to pass it along to the resource. It's self-signed, so you should be able to use it as a CA.

commands/out.go
Infra-Red: I updated the `out` command to use a method from `Source`.

commands/out.go
```diff
@@ -248,8 +258,15 @@ func createAuth(req resource.OutRequest) *authn.Basic {
 func aliasesToBump(req resource.OutRequest, repo name.Repository, ver *semver.Version) ([]name.Tag, error) {
 	variant := req.Source.Variant
 
-	opts := []remote.Option{}
-	opts = append(opts, remote.WithAuth(createAuth(req)))
```
vito, line 252: Can this `createAuth` function be removed now?

vito (changes requested):
Thanks, this looks great! Just nits to pick from here. I do think those new `It`s are a bit too noisy, but thanks for the due diligence. 🙂

commands/out.go
Infra-Red: `createAuth ` is still used in the `signImages` function. The `gcr.NewTrustedGcrRepository()` function requires authorization to be passed.

check_test.go
```diff
@@ -67,6 +70,10 @@ var _ = Describe("Check", func() {
 				req.Version = nil
 			})
 
+			It("should not error", func() {
```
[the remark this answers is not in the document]
Infra-Red: I added this assertion to existing `It` blocks, I think this is the most common pattern. Fixed in https://github.com/concourse/registry-image-resource/pull/264/commits/40734d20866202465015bbdc37c337542cc5ae2c.

in_test.go
```diff
@@ -86,6 +90,10 @@ var _ = Describe("In", func() {
 			}
 		})
 
+		It("should not error", func() {
```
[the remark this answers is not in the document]
Infra-Red: Fixed in https://github.com/concourse/registry-image-resource/pull/264/commits/40734d20866202465015bbdc37c337542cc5ae2c.

vito (approved):
LGTM - thanks again!

PR #276, opened 2021-04-26
allow setting a semver constraint

aoldershaw:
Semver constraints allow for constraining the allowed versions. e.g. to pin to a minor version, you can use `semver_constraint: 1.2.x`.

### DIFF
--- a/commands/check.go
+++ b/commands/check.go
@@ -129,6 +129,14 @@ func checkRepository(repo name.Repository, source resource.Source, from *resourc
 		})
 	}
 
+	var constraint *semver.Constraints
+	if source.SemverConstraint != "" {
+		constraint, err = semver.NewConstraint(source.SemverConstraint)
+		if err != nil {
+			return resource.CheckResponse{}, fmt.Errorf("parse semver constraint: %w", err)
+		}
+	}
+
 	for _, identifier := range tags {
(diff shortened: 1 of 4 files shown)


commands/check.go
```diff
@@ -246,7 +259,7 @@ func checkRepository(repo name.Repository, source resource.Source, from *resourc
 		digest := tagDigests[latestTag]
 
 		_, existsAsSemver := digestVersions[digest]
-		if !existsAsSemver {
+		if !existsAsSemver && constraint == nil {
```
chenbh, line 262: If I'm understanding this correctly, this will ensure that if I configure the resource with a constraint and a variant (`latest`), it will only return the `("latest", digest)` combo if there's another `(tag, digest)` where the `tag` satisfies the semver constraints?

aoldershaw: If you specify a semver constraint, it'll never give you the `latest` (or bare variant) tag. The existing default behaviour is to only emit the `latest` tag if there is no semver tag with the same digest. When you specify a semver constraint, though, the tags that are returned must satisfy that constraint - if `latest` doesn't correspond with a semver tag, then it certainly doesn't satisfy that constraint

chenbh (approved):

PR #282, opened 2021-06-06
Optionally allow out version to be specified in a file

Chraneco:
This will allow pushing a tagged image to be combined with the semver resource, for example.


vito:
I would prefer that folks use the new [`load_var` step](https://concourse-ci.org/jobs.html#load-var-step) for this instead, so that resource types don't have to maintain support for files in addition to static values.

```yaml
plan:
- get: version
- load_var: version_tag
  file: version/version
- put: image
  params: {version: ((version_tag))}
```

Thanks for the PR and tests though!

PR #288, opened 2021-07-08
Add support for AWS IAM roles chain

fredericfran-gds:
In certain scenarios, a shared Concourse is run in a separate AWS
account (Account C) and needs to assume a role in a different AWS
account (Account A) in order to do the concourse tasks. The ECR is
located in another AWS account: Account E. Only Account A has access to
Account E. Therefore for registry-image to access ECR using the Account
C credentials, it needs to assume a role in Account A; which allows
assuming a role in Account E to have access to the ECR.

This PR introduces a new parameter: `aws_role_arns` which takes a an
array of IAM roles to be assumed in the order specified.

Users can still use the previous `aws_role_arn` that we kept for
backward compatibility but if both are specified, an error will be
thrown by the resource.

The following local test were performed locally:
1. iam roles chain (new feature)
2. error if both `aws_role_arn` and `aws_role_arns` are set
2. only `aws_role_arn` specified (backward compatibility)

Signed-off-by: Frederic Francois <frederic.francois@digital.cabinet-office.gov.uk>

### DIFF
--- a/README.md
+++ b/README.md
@@ -74,7 +74,12 @@ differences:
 * `aws_region`: *Optional. Default `""`.* The region to use for accessing ECR. This is required if you are using ECR. This region will help determine the full repository URL you are accessing (e.g., `012345678910.dkr.ecr.us-east-1.amazonaws.com`)
 
 * `aws_role_arn`: *Optional. Default `""`.* If set, then this role will be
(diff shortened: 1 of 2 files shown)


aoldershaw (approved):
LGTM! Thanks for making the changes!

PR #289, opened 2021-07-09
Enable GCP Authentication 

### LINKED ISSUE #280: Google Container Registry authentication via oauth2 token
It'd be nice to be able to rely on a Concourse worker's IAM roles in GCP to authenticate with Google Container Registry.

Step 5 of https://cloud.google.com/container-registry/docs/advanced-authentication#token shows how to log in to regular docker using a token obtained from Google's metadata servers (e.g. workload identity from a VM or K8s container).

The Google container libraries that this project uses should already support this, but I think we'd need to avoid the checks for username/password or require a special-case username to allow the library to retrieve credentials.

A manual way, from a GCP instance / GKE container (without e.g. jq) would be:

```
curl -H "Metadata-Flavor: Google" http://169.254.169.254/computeMetadata/v1/instance/service-accounts/default/token \
| sed 's/.*access_token":"\(.*\)","expires.*/\1/' \
| docker login -u oauth2accesstoken --password-stdin https://eu.gcr.io/some-repo
```

Thoughts?

reedk-au:
Fixes #280 

Add a flag to source that will look for and use GCP application default credentials instead of using AWS credentials 0

### DIFF
--- a/types.go
+++ b/types.go
@@ -97,7 +97,7 @@ type Source struct {
 
 	Debug bool `json:"debug,omitempty"`
 
-	GcpAuth bool `json:"gcp_auth,omitempty"`
+	GcpTokenSource *string `json:"gcp_token_source,omitempty"`
 }
 
 func (source Source) Mirror() (Source, bool, error) {
@@ -280,11 +280,11 @@ func (source *Source) Metadata() []MetadataField {
 	}
 }
 
-func (source *Source) AuthenticateToGCP() bool {
+func (source *Source) AuthenticateToGCP(tokenSource string) bool {
 	logrus.Warnln("GCP integration is experimental and untested")
 	logrus.Info("Using GCP service account credentials to authenticate to GCP")
 
-	oauthToken := google.ComputeTokenSource("")
+	oauthToken := google.ComputeTokenSource(tokenSource)
 	token, err := oauthToken.Token()
 
 	if err != nil || token == nil {
--- a/types.go
+++ b/types.go
@@ -24,6 +24,7 @@ import (
 	"github.com/google/go-containerregistry/pkg/v1/remote"
 	"github.com/google/go-containerregistry/pkg/v1/remote/transport"
 	"github.com/sirupsen/logrus"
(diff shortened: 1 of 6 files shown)


reedk-au:
Implementing the same on `in` and `out`, moved into draft status until completed. Comments and recommendations welcome in the meantime.

chenbh:
Hey, sorry for the late response.

I don't think this PR can be accepted in the current form, the reasoning is the same as the one in https://github.com/concourse/concourse/issues/3023 and Aidan's comment in https://github.com/concourse/registry-image-resource/pull/287#issuecomment-876483708. 

The tldr is that Concourse differentiates resources based on their `source`, and automagic metadata servers breaks this assumption. With this PR, if I have 2 worker nodes with different credentials, it is now possible for a `check` step to return different versions depending on which worker it was run on (if each VM's creds can see different GCR repos).

As an alternative, I would be open to something like the current AWS flow, where the authenticated session is constructed from credentials passed in from the `source`:
https://github.com/concourse/registry-image-resource/blob/4372334d66d137cf536ef5c1ce2cd6944b0c5506/types.go#L288-L291

Maybe you can do something similar by using [`google.CredentialsFromJSON`](https://pkg.go.dev/golang.org/x/oauth2/google#CredentialsFromJSON) and taking the `TokenSource` from there?

types.go
```diff
@@ -95,6 +96,8 @@ type Source struct {
 	DomainCerts []string `json:"ca_certs,omitempty"`
 
 	Debug bool `json:"debug,omitempty"`
+
+	GcpTokenSource *string `json:"gcp_token_source,omitempty"`
```
chenbh, line 100: This would also need a corresponding README change

reedk-au:
No worries, thanks for the review @chenbh and this makes sense. Will see if we can get this to work and move this PR to a draft in the meantime.

michaelpporter:
What is needed to get this active again?

taylorsilva:
Closing because stale.
Anyone is free to pick this up and re-submit the PR.

Issue #281, opened 2021-05-24
Error when using `bump_aliases: true` for first image push


pn-santos:
I think I found an issue when using `bump_aliases: true` for images that do not yet exist in the repository.

### Scenario

**Given** a semver-based (no tag) `registry-image` resource, e.g.:
```yaml
  - name: jre.oci
    type: registry-image
    icon: docker
    source:
      repository: my.repo.com/runtime/jre
    check_every: never
```

**And** Image in question doesn't yet exist in the repo

**When** a `put` specifying `bump_aliases` is done:
```yaml
- put: jre.oci
  inputs: detect
  params:
    image: image/image.tar
    version: ((.:image-version))
    bump_aliases: true
  get_params:
     skip_download: true
```

**Then** the image should be pushed with `latest` and all `M`, `M.m` and `M.m.p` versions

### What actually happens

The `put` step logs this error:
```
ERRO[0000] determine aliases: list repository tags: GET https://my.repo.com/v2/runtime/jre/tags/list?n=1000: NAME_UNKNOWN: repository name not known to registry; map[name:runtime/jre]
```

I tried manually pushing the image and re-running the job and then it worked correctly.

### Versions

- Concourse v7.2.0
- `registry-image` resource that was bundled with v7.2.0

tremblaysimon:
I reproduce the same problem here with registry-image-resource version 1.4.0 and using Artifactory as Docker Registry.

PR #291, opened 2021-07-12
Fix semver first image push

pn-santos:
Fixes https://github.com/concourse/registry-image-resource/issues/281

There's a couple of things I'm not too sure about:

1. Should the code just check for 404 status code and ignore the response content? I chose to also parse the content to attempt to distinguish between a 404 due to a request not hitting the registry and a 404 due to an image not existing.

2. I tried to come up with a test that would "fit" the existing structure, it works but if there's a "nicer" way to go about I welcome any suggestion 

3. The reason I'm parsing the `NAME_UNKNOWN` and `NOT_FOUND` is because the first one is what https://github.com/distribution/distribution returns and the second one is what https://github.com/goharbor/harbor returns (so I think this might vary between registry implementations?)

**note:** Not a go dev so code-wise YMMV (but I think it's mostly ok?)

### DIFF
--- a/commands/out.go
+++ b/commands/out.go
@@ -4,6 +4,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"io"
+	"net/http"
 	"os"
 	"path/filepath"
 	"strings"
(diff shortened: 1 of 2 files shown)


pn-santos:
@xtremerui I don't want to be too annoying but any chance this could be looked at soon-ish? It's a somewhat annoying issue for new images and (I hope) fairly straightforward fix (putting aside minor implementation details I listed in the OP)

xtremerui (approved):
LGTM

xtremerui:
Sorry I somehow missed the notification of reviewing this PR. Thx for the reminder.

Please feel free to pin the assignee if the PR got assigned and being no activity for at most 2 weeks.

PR #295, opened 2021-07-30
Include labels in metadata.json

martinxsliu:
Hi all, this small change adds the image's labels to the `metadata.json` file that's produced for the `rootfs` format. This is useful for us as we can include things like VCS metadata in the labels which can be used in tasks.

### DIFF
--- a/commands/in.go
+++ b/commands/in.go
@@ -19,8 +19,9 @@ import (
 )
 
 type ImageMetadata struct {
-	Env  []string `json:"env"`
-	User string   `json:"user"`
+	Env    []string          `json:"env"`
+	User   string            `json:"user"`
+	Labels map[string]string `json:"labels"`
 }
 
 type In struct {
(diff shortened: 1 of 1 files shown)


aoldershaw:
Hi @martinxsliu - sorry for the late response, the team had the past week off.

Is there a reason you want the labels in `metadata.json` in particular? I ask because `metadata.json` is the file that's used by Concourse specifically to set the environment vars and user when running the container - Concourse itself would have no use for labels information, at least currently.

How would you feel about writing this content to a separate file, e.g. `labels.json`, like we do for `tag`, `digest`, and `repository`: https://github.com/concourse/registry-image-resource/blob/6876ee5cf787a02918eee0ecbb36f36d3da7d49e/commands/in.go#L171-L185

mihneastaub:
Hi Guys,

I see that for the `registry-image-resource` you cannot take the labels from the image, from what I understand from the README file. And from what I see this pr will do this think, do you know when this pr will be merged?

xtremerui:
@mihneastaub it depends on if @martinxsliu gonna take the suggestion from comment and update this PR

taylorsilva:
PR not updated in over a year

PR #297, opened 2021-08-15
Add AWS Instance Profile authentication support for AWS ECR

jimmyR:
Following AWS Best Practice, we use Instance Profiles on our EC2 nodes that run concourse.

By avoiding passing Credentials to the aws.Config, it allows the registry-image-resource to use Instance Profiles authentication, so we can avoid having to create and manages access keys.

### DIFF
--- a/README.md
+++ b/README.md
@@ -63,10 +63,10 @@ differences:
   using `put`.
 
 * `aws_access_key_id`: *Optional. Default `""`.* The access key ID to use for
-  authenticating with ECR.
+  authenticating with ECR.  When using Instance Profiles this can be left blank.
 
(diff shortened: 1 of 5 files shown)


aoldershaw:
Hi, thanks for the PR, but unfortunately I don't think we'll accept it. For some context as to why, refer to https://github.com/concourse/registry-image-resource/pull/287#issuecomment-876483708, but to try and summarize:

* This breaks one of Concourse's core assumptions (that the `source` is the actual source of truth of a resource)
* We have an alternative fix in the works that should allow you to use instance profiles while still making the `source` explicit

If you need this behaviour for the time being, and are okay with the implications outlined in the linked comment, than you can push an image based on your fork and use that as a custom `resource_type`

PR #305, opened 2021-11-04
lets see if we can break the tests

rockhowse:
breaking and fixing the attempts at using "insecure" for the `registry-image` concourse resource.


rockhowse:
whooops meant to open this against my fork... closing it out =P

Issue #277, opened 2021-05-05
Clarity on aws_role_arn


luis-nagaki:
I am assuming that the aws_role_arn is NOT assuming the role (bc the ec2 instance has access to assume it) and should be able to go and fetch the docker image from ecr without having to add the access key and secret.

I am trying to just use
resources:

    name: myrepo
    type: registry-image
    source:
    repository: '12345678.dkr.ecr.us-west-1.amazonaws.com/mytestrepo'
    aws_role_arn: 'arn:aws:iam::12345678:role/testrole_has_access'
    aws_regionL 'us-west-1'

I can successfully do any pull or push via aws cli and assuming this role within the workers, but when this resources tries i get failed.. ERRo[0000] checking origin 12345678.dkr.ecr.us-west-1.amazonaws.com/mytestrepo failed: list repository tags: GET https://12345678.dkr.ecr.us-west-1.amazonaws.com/mytestrepo/tags/list?n=1000L unexpected status code 401 Unauthorized: Not authorized

fyi, i gave all perms to the role - even getauthtoken. all this works via cli but not this resource. (If thats what aws_role_arn is used for)

Concourse 7.2.0

luis-nagaki:
Read a lot regarding security and etc in past open issues, but no final thoughts on where it's going. Are there any updates regarding the use of aws_role_arn for assuming the role?

jorymorrison:
+1

samed:
> Leaving a message here to let folks know that the core team doesn't use AWS so we're quite out of our depth here. Help from someone in the community who can troubleshoot this against AWS would be great. We'd be happy to help from the code side, giving pointers and showing how to run thing if needed.

@taylorsilva I've managed to get it working by just relying on `req.Source.AwsRegion` in the source, so in this way, I made it fallback to instance IAM profile without using hardcoded was credentials. AWS-SDK has the following precedence (https://docs.aws.amazon.com/sdk-for-go/api/aws/session/) for credentials/sessions:

* Environment Variables
* Shared Credentials file
* Shared Configuration file (if SharedConfig is enabled)
* EC2 Instance Metadata (credentials only)




Here's my patch, which is just an (ugly) workaround. 


```
diff --git i/commands/check.go w/commands/check.go
index fa93368..72512d1 100644
--- i/commands/check.go
+++ w/commands/check.go
@@ -49,7 +49,7 @@ func (c *Check) Execute() error {
                return fmt.Errorf("invalid payload: %s", err)
        }

[...]

PR #309, opened 2021-12-16
Add aws_use_instance_role flag

jprelph:
This is in relation to the bug discussed here: https://github.com/concourse/registry-image-resource/issues/277

The current default behaviour doesn't allow EC2 IAM Roles to be assumed directly, I have added an aws_use_instance_role flag which - when set to true alongside the instance name - ignores other cred settings when authenticating with ECR to allow the instance role to be assumed.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 testdata/.push-output
 .envrc
 tags
+.idea
--- a/README.md
+++ b/README.md
@@ -83,7 +83,7 @@ differences:
   If set, these roles will be assumed in the specified order before authenticating to ECR.
   An error will occur if `aws_role_arn` is also specified.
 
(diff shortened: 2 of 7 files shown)


xtremerui:
Thanks for the PR. Please refer to https://github.com/concourse/registry-image-resource/pull/287#issuecomment-876483708.

jprelph:
> Thanks for the PR. Please refer to [#287 (comment)](https://github.com/concourse/registry-image-resource/pull/287#issuecomment-876483708).

Thanks - I'd not seen that in relation to this, but having a look through that I'm not sure why the particular risk in terms of using the metadata service here would be. This isn't a case of using the EC2 IAM role to pull in some credentials that then are templated in anywhere (ie. by passing back a secret access key/access key id which we then plug in)- it's relying on the behaviour of the aws-sdk to generate a session token as part of the call to the aws service. I don't see fundamentally what's different from the workers being capable of having different configuration (ie. different DNS servers, env vars etc.) and/or how prototypes really changes the risk there?

taylorsilva:
Closing this issue and it won't be merged

PR #321, opened 2022-06-18
Add support for building OCI multi-arch images

rdrgmnzs:
Add support for building OCI multi-arch images.

It should work in conjunction with the work on https://github.com/concourse/oci-build-task/pull/93

This PR assumes that the image is being provided to it as a directory rather than a tar. This is because google/go-containerregistry does not support importing multi-arch images at this time. oci-build-task is already extracting the image to a directory for similar reasons, so instead of doubling up the work we just consume the directory it creates.

Signed-off-by: Rodrigo Menezes <brdude@gmail.com>


commands/out.go
```diff
@@ -187,42 +207,74 @@ func (o *Out) Execute() error {
...
 	if req.Source.ContentTrust != nil {
-		err = signImages(req, img, tags)
-		if err != nil {
-			return fmt.Errorf("signing image(s): %w", err)
+		switch t := img.(type) {
+		case v1.Image:
+			err = signImages(req, t, tags)
+			if err != nil {
+				return fmt.Errorf("signing image(s): %w", err)
+			}
```
xtremerui, line 232: what if it is v1.ImageIndex here?

xtremerui (commented):
Thx for the PR! We will need to add tests to https://github.com/concourse/registry-image-resource/blob/7a71c3deca00cbd47f67ec8ee76800231293ddb9/out_test.go since currently the tests assume only image but not image index.

commands/out.go
```diff
@@ -58,7 +60,7 @@ func (o *Out) Execute() error {
 	}
 
 	if len(o.args) < 2 {
-		return fmt.Errorf("destination path not specified")
+		return fmt.Errorf("import path not specified")
```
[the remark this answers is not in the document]
rdrgmnzs: Truthfully not sure, reverted back.

commands/out.go
rdrgmnzs: It'll fall to default and fail with an error `cannot sign type v1.ImageIndex`, unfortunately github.com/simonshyu/notary-gcr doesn't support signing v1.ImageIndex at this time.

xtremerui: @evanchaoli ^ FYI

rdrgmnzs:
@xtremerui added some tests which should determine if the image is pushed correctly. I decided against duplicating the glob matching tests as I don't think there'd be a lot to gain from them but let me know if you want me to add them anyway.

scottwallacesh:
I am trying to build two images, one for `amd64` and one for `arm64`, tag them with their version (e.g. `1.2.1`) and produce a single `latest` tag/image/manifest that includes both the architectures.

I've tried this branch/PR but am getting:
```
could not load image from path 'image/image.tar': file manifest.json not found in tar 
```

Job::
```yaml
jobs:
  - name: build-test-and-push
    plan:
      - get: sourcecode
        trigger: true
      - get: version
        params:
          bump: patch
      - get: tags
      - load_var: image-version
        file: version/version
      - get: concourse-pipelines
      - task: build-image
        privileged: true
        file: concourse-pipelines/tasks/build-image.yaml
      - task: run-tests
        privileged: true
        file: concourse-pipelines/tasks/run-tests.yaml
      - put: remote-image
        params:
          image: image/image.tar
          version: ((.:image-version))
          bump_aliases: true
      - put: version
        params:
          file: version/version
```

`build-image.yaml`:
```yaml
platform: linux
image_resource:
  type: registry-image
  source:
    repository: nexus.example.com:8449/oci-build-task  # image built from this branch/PR
    tag: multiarch
inputs:
  - name: sourcecode
outputs:
  - name: image
  - name: cache
params:
  OUTPUT_OCI: true
  CONTEXT: sourcecode
  REGISTRY_MIRRORS: nexus.example.com:8443
  IMAGE_PLATFORM: linux/amd64,linux/arm64
run:
  path: build
```

What have I missed/misunderstood?

commands/out.go
evanchaoli: @SimonXming ^ FYI

xtremerui (approved):
LGTM

PR #323, opened 2022-07-19
Relax auth around ECR

### LINKED ISSUE #277: Clarity on aws_role_arn

taylorsilva:
Fixes #277

Kudo's to @samed and his comment here: https://github.com/concourse/registry-image-resource/issues/277#issuecomment-960285967

Basically the same patch with an added conditional for optionally adding creds if they're present.

I've tested this on a Concourse worker running on AWS with an instance profile that has the relevant ECR permissions. `check`, `get`, and `put` steps all work.

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,7 +61,9 @@ differences:
     <br>
     <em><strong>Note:</strong> If using ecr you only need the repository name,
     not the full URI e.g. <code>alpine</code> not
-    <code>012345678910.dkr.ecr.us-east-1.amazonaws.com/alpine</code></em>
+    <code>012345678910.dkr.ecr.us-east-1.amazonaws.com/alpine</code>. ECR usage
+    is NOT automatically detected. You must set the <code>aws_region</code> to
+    tell the resource to automatically use ECR.</em>
     </td>
   </tr>
   <tr>
--- a/commands/check.go
+++ b/commands/check.go
@@ -49,7 +49,7 @@ func (c *Check) Execute() error {
 		return fmt.Errorf("invalid payload: %s", err)
 	}
 
-	if req.Source.AwsAccessKeyId != "" && req.Source.AwsSecretAccessKey != "" && req.Source.AwsRegion != "" {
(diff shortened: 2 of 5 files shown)


taylorsilva:
One "gotcha" I thought of is that with this change I'm assuming that anyone that has `aws_region` WANTS to use ECR authentication. I have a feeling that may not always be true, or at least don't have any evidence to tell me otherwise. Going to make a change, probably adding another field, so ECR auth will trigger when either key ID/secret are set OR a new `aws_auth` is set.

taylorsilva:
Looking at the before and after of the code I realized that no one would have set `aws_region` and NOT also set the key ID/secret. Currently if you set only `aws_region` then the resource does nothing different and you'll get an auth error. This is also true even if your worker is not running inside aws infrastructure. Therefore I don't think this is a breaking change at all. There's no scenario where someone has JUST `aws_region` and wouldn't want to also authenticate to ecr.

xtremerui (approved):
LGTM!

PR #324, opened 2022-08-16
Uid&Gid info lost for emptied directory while unpack image

SimonXming:
If a emptied directory created in Dockerfile, it will not `chown` to restore the origin uid&gid during unpack image. This was handled by `tarfs.ExtractEntry`
- https://github.com/vmware-archive/go-archive/blob/v1.0.1/tarfs/extract.go#L123-L128
- introduced since https://github.com/concourse/registry-image-resource/pull/284.

Signed-off-by: Ming Xu <mingx@vmware.com>

### DIFF
--- a/commands/unpack.go
+++ b/commands/unpack.go
@@ -119,7 +119,7 @@ func extractLayer(dest string, layer v1.Layer, bar *mpb.Bar, chown bool) error {
 			if err := os.Mkdir(dir, fi.Mode()&os.ModePerm); err != nil {
 				return err
 			}
-			// Fix issue introduced by https://github.com/concourse/registry-image-resource/pull/284
+
 			if runtime.GOOS != "windows" && chown {
 				err = os.Lchown(dir, hdr.Uid, hdr.Gid)
 				if err != nil {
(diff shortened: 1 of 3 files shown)


SimonXming:
> There is test failure on both checks, you might want to fix those
> 
I think it's because `testdata/whiteout/Dockerfile` get changed in the PR (adding new directory `rootfs/top-dir-4`), but the  `concourse/test-image-whiteout` image which used in the unit test not updated with modified Dockerfile in the PR. These two failure should be fixed once `concourse/test-image-whiteout` image get updated.

commands/unpack.go
```diff
@@ -118,7 +119,13 @@ func extractLayer(dest string, layer v1.Layer, bar *mpb.Bar, chown bool) error {
 			if err := os.Mkdir(dir, fi.Mode()&os.ModePerm); err != nil {
 				return err
 			}
-
+			// Fix issue introduced by https://github.com/concourse/registry-image-resource/pull/284
```
xtremerui, line 122: We dont really need to add comment here for the fix. Instead we should make it as detail as possible in this PR thread(commit message and PR description).

For example you could add some details about how the linked PR introduced the bug e.g. removing a dir and adding it back without owner info (feel free to use whatever you think that'd be more clear).

Thx!

SimonXming: Sure, I will remove this comment.

xtremerui (approved):
LGTM!

Issue #327, opened 2022-11-17
Check returns version without digest


xtremerui:
This behaviour has being only found in some resources with type `registry-image-resource` accross production CI. Could be a glitch of docker API when `check` querying for version info. When there is newer version other than this problematic one available, the `check` will work again most likely.

<img width="628" alt="Screen Shot 2022-11-16 at 9 27 18 PM" src="https://user-images.githubusercontent.com/1585949/202339603-3977cffa-0157-4eef-909e-b3a54ef20469.png">

And when this happens, the get step of this resource will fail by error
```
selected worker: ci-workers-worker-2
fetching aquasec/trivy@
ERRO[0000] download failed: get image: GET https://index.docker.io/v2/aquasec/trivy/manifests/: unexpected status code 404 Not Found: 404 page not found 
```
since no digest was provided for the GET request.

The temp workaround is to disable that bad version and rerun the build OR switching to docker-image resource.

xtremerui:
Further observation is this could also affect a resource type or image resoure that with `type` to be registry-image-resource. When the check again returns a version without digest, for any step that runs on the image, it will errored by
```
image fetching failed
``` 
when the step is trying to initilize the container by first pulling the base image for it.

ramonskie:
im experiening the same problem since yesterday
the pipeline has not been updated in a while so it seems a change within docker.io
```
- name: metalink-repository
  source:
    repository: dpb587/metalink-repository-resource
  type: registry-image
```
```
fetching dpb587/metalink-repository-resource@
ERRO[0000] download failed: get image: GET https://index.docker.io/v2/dpb587/metalink-repository-resource/manifests/: unexpected status code 404 Not Found: 404 page not found 
```

FelixMarxIBM:
We noticed the same issue in our environment and this only happened if no tag is present as some of our resources still worked

Doesn't work and shows the same issue
```
resource_types:
  - name: google-cloud-storage
    type: registry-image
    source:
      repository: frodenas/gcs-resource
```

However the following resource type does work and is able to pull the latest tag.
This means that the tag is not **optional** anymore when used with registry-image.
```
resource_types:
  - name: google-cloud-storage
    type: registry-image
    source:
      repository: frodenas/gcs-resource
      tag: latest
```

xtremerui:
thx for the update. Tag will sure filter out the wrong version. there is potential some logic diff for querying docker with or without tag. I will look into it.

jpalermo:
[...]

PR #328, opened 2022-11-18
Try to fix malform version for check command

xtremerui:
This PR bumps go-containerregistry and adds a nil check before returning version digest, in hoping to fix https://github.com/concourse/registry-image-resource/issues/327.

The latest go-containerregistry seems reduce un-necessary API call so I can clean up some tests, which should help rate-limit problem.


xtremerui:
will merge this into master and test the latest concourse/registry-image:dev in production CI

PR #332, opened 2023-01-09
Clarify tags included with `additional_tags`

imsodin:
The default tag `default` isn't pushed when using the `additional_tags` option.

Maybe overly verbouse? It would have helped me though :)


taylorsilva (approved):
LGTM

Issue #329, opened 2022-11-29
Support for Signing Images with Cosign


ChrisJBurns:
We are using Harbor to host our images and we're noticing that the newer versions of Harbor now won't be supporting Notary for their image signing methods and are aligning more with the Cosign toolstack.

Can we get support for signing of images with Cosign?

ChrisJBurns:
For those reading, we've had an increasing need for this functionality for our own internal project as we are using Cosign and having to do this in separate tasks is becoming increasingly tricky to do due to several awkward factors. I've got something to work and will be using this internally as a trial and if everything looks go with no issues, I will raise the PR for this repository and hopefully others can benefit from this.

Changes can be found here (subject to change if we need to tweak certain bits): https://github.com/concourse/registry-image-resource/compare/master...ChrisJBurns:registry-image-resource:adds-cosign-support?expand=1

theabrar:
any update on this? as we plan to use cosign as well, thanks.

taylorsilva:
I closed the related PR because it's stale. If anyone has bandwidth to work on this, feel free to re-submit the PR.

I wonder if the library support is any better for this now and would make implementing easier?
https://github.com/sigstore/cosign/tree/main/pkg/cosign
https://pkg.go.dev/github.com/sigstore/cosign/v2/pkg/cosign

PR #341, opened 2023-06-06
feat: adds cosign support

ChrisJBurns:
Currently, Content Trust is a feature offered by the `registry-image-resource`, but the gold standard for image signing now is using (if it isn't already) Cosign. This PR implements the functionality to sign images with Cosign. Currently, although it is possible using `SignCmd`, Cosign is a CLI first - meaning using it as a library in code can be a little less than ideal. This results in large number of config objects having to be passed into the `SignCmd` due to the fact that there is no CLI framework setting the default values. 

When Cosign becomes more and more usable as a library, the code in this PR can be reduced. This includes, the way we have to set a temporary environment variable for `COSIGN_KEY` and `COSIGN_PASSWORD` until these are values that can be more easily passed into the Cosign code. Another one is the Keychain. Currently, Cosign works that if you have a docker config JSON file with registries and auth configured for them in a local cred store, Cosign will just use them via the `go-containerregistry` libary. Due to security reasons, we don't want to have to put the credentials in a file in the `registry-image-resource` task as any developer that intercepts the container can easily view those credentials. Instead we use an InMemoryKeychain that the underlying Cosign/go-containerregsitry libraries will pick up and use for the pushing of signatures to the registry.

implements: https://github.com/concourse/registry-image-resource/issues/329


taylorsilva (changes requested):
Thanks for the PR Chris! Sorry for the long delay in taking a look at it.

I'm noticing there are no tests for this feature. We'd need some tests before merging this in: https://github.com/concourse/registry-image-resource/blob/master/out_test.go

ChrisJBurns:
Thanks @taylorsilva , I'll get those added!

ChrisJBurns:
@taylorsilva 

I've added a test but I'm not entirely sure if there's a way to test if the time has been signed as there is no response, only an error. So I've just asserted that there is no error that comes back for that test. Wondering what other thoughts you had?

commands/out.go
```diff
@@ -238,6 +250,158 @@ func put(req resource.OutRequest, img partial.WithRawManifest, tags []name.Tag,
...
+	// image digests that stay the same all the time, this is where tools like `apko`
+	// are a great tool for building OCI images)
+	imgDigestUrl := tags[0].String() + "@" + digest.String()
+
+	// over time this won't be needed. the Cosign library will evolve to expect
+	// less object parameters to be configured. this is only needed for now due to
+	// the fact that primarily cosign is a cli tool. And part of the CLI framework
+	// that cosign uses, sets the majority of the below objects variables by default.
+	// this comment concerns the below `ro`, `o` and `ko` data objects.
+	ro := &options.RootOptions{
```
taylorsilva, line 304: nit:
```suggestion
	// this comment concerns the below `rootOpts`, `signOpts` and `keyOpts` data objects.
	rootOpts := &options.RootOptions{
```

commands/out.go
```diff
@@ -238,6 +250,158 @@ func put(req resource.OutRequest, img partial.WithRawManifest, tags []name.Tag,
...
+	if err != nil {
+		return fmt.Errorf("err %w", err)
+	}
+
+	// similiar to the COSIGN_KEY variable we set the password that was used to create the
+	// keypair (if there was one). there are less ways to set the password currently outside
+	// of an environment variable or user input via a terminal prompt, as the Cosign library
+	// evolves over time, we can reasonably expect this to change, but as we cannot rely on user
+	// input, we have to use the environment variable.
+	err = os.Setenv("COSIGN_PASSWORD", req.Source.Cosign.Password)
```
taylorsilva, line 391: Should we verify that `Cosign.Key` and `Cosign.Password` are set? Is it okay of they're blank? Are both/one required?

commands/out.go
```diff
@@ -238,6 +250,158 @@ func put(req resource.OutRequest, img partial.WithRawManifest, tags []name.Tag,
...
+	if err != nil {
+		return fmt.Errorf("err %w", err)
+	}
+
+	logrus.Infof("Signing image with Cosign: %s", imgDigestUrl)
+	err = sign.SignCmd(ro, ko, o, []string{imgDigestUrl})
+	if err != nil {
+		return fmt.Errorf("there was an error signing the image with Cosign %w", err)
```
taylorsilva, line 399: nit: add a colon to separate our error message and the upstream error message
```suggestion
		return fmt.Errorf("there was an error signing the image with Cosign: %w", err)
```

out_test.go
```diff
@@ -287,6 +287,23 @@ var _ = Describe("Out", func() {
...
+					Registry: "test.registry",
+					Key:      "",
+					Password: "",
+				}
+			})
+
+			It("Signs the image", func() {
+				// there is little to know if the image has been signed
+				// so we assume that if there is no error, then it has been signed
+				Expect(actualErr).ToNot(HaveOccurred())
```
taylorsilva, line 303: We can check the output from the command contains the `logrus.Infof()` message, could we not?
```suggestion
			It("Signs the image", func() {
			Expect(actualErrOutput).To(ContainSubstring("Signing image with Cosign"))
			Expect(actualErrOutput).To(ContainSubstring("Image signed with Cosign"))
				Expect(actualErr).ToNot(HaveOccurred())
```

commands/out.go
taylorsilva: Based on your test it looks like only the registry url is required. Is that correct?

taylorsilva (changes requested):
Overall this is looking good. I made a suggestion for the test, let me know what you think.

I left a bunch of code-style related comments. My intention behind those comments is to keep the coding style within the code-base consistent.

I noticed the `README` hasn't been updated with the new `cosign` field. Could you please update the table with a new row describing which fields in `cosign` are required/optional. Not sure if the `cosign` field should live in the main `Source` or in `PutParams` instead. I left a comment below about that as well. Would someone want to use `cosign` in the `get` step? If they would, then it makes sense to leave `Cosign` in the `Source` struct.

Thanks again for the PR and making the changes you've made so far! Next review will go faster now that I've actually read all the code :)

types.go
```diff
@@ -105,6 +105,8 @@ type Source struct {
 
 	ContentTrust *ContentTrust `json:"content_trust,omitempty"`
 
+	Cosign *Cosign `json:"cosign,omitempty"`
```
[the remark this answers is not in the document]
ChrisJBurns: @taylorsilva Is an interesting point on the `get`. Wondering if I could do that in a separate PR? Correct me if I'm wrong, but this would allow us to verify that the `registry-image-resource` image that we are pulling is signed?

commands/out.go
```diff
@@ -238,6 +250,158 @@ func put(req resource.OutRequest, img partial.WithRawManifest, tags []name.Tag,
...
+type userCredentials struct {
+	username string
+	password string
+}
+
+// A keychain that is able to return the creds based on the registry url
+type InMemoryKeyChain struct {
+	// An in-memory map of username/passwords
+	credentials map[string]userCredentials
+}
```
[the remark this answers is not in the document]
ChrisJBurns: Yep good point. Have committed the changes.

commands/out.go
```diff
@@ -238,6 +250,158 @@ func put(req resource.OutRequest, img partial.WithRawManifest, tags []name.Tag,
...
+
+// Returns a function that is able to produce an AuthConfig struct. This is basically a factory factory
```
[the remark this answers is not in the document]
ChrisJBurns: Yep good point. Have committed the changes.

commands/out.go
```diff
@@ -238,6 +250,158 @@ func put(req resource.OutRequest, img partial.WithRawManifest, tags []name.Tag,
...
+func (k *InMemoryKeyChain) Resolve(resource authn.Resource) (authn.Authenticator, error) {
+
+	// Ask for the registry URL
```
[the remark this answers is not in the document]
ChrisJBurns: :+1:

commands/out.go
```diff
@@ -238,6 +250,158 @@ func put(req resource.OutRequest, img partial.WithRawManifest, tags []name.Tag,
...
+	registryHost := resource.RegistryStr()
+
+	// Find the user credentials by the registry URL
```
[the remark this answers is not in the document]
ChrisJBurns: :+1:

commands/out.go
```diff
@@ -238,6 +250,158 @@ func put(req resource.OutRequest, img partial.WithRawManifest, tags []name.Tag,
...
+	userCreds, ok := k.credentials[registryHost]
+
+	if !ok {
+		return authn.Anonymous, fmt.Errorf("unable to find credentials for host %s", registryHost)
+	}
+
+	// Return an authConfig that contains the username and password we looked up
```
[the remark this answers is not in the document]
ChrisJBurns: :+1:

commands/out.go
ChrisJBurns: :+1:

out_test.go
ChrisJBurns: Have added those lines in, although interestingly, even with the cosign key and password being `""` in the test, there are no failures.

commands/out.go
ChrisJBurns: :+1:

commands/out.go
ChrisJBurns: Yep, have added some checks :+1:

taylorsilva:
The `out` test is failing:
image

Slightly more detailed error:
image

types.go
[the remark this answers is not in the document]
taylorsilva: Yes, it would! and totally fine to do that in another PR. Just wanted to determine if it made sense to keep the `cosign` field in the top-level source, or bring it down to the `params` used by the `out` command. Sounds like it makes sense to keep it up in `source` so someone can use it in a `get` step in a future PR.

taylorsilva:
Made a small fix to the test. Looks like the image isn't being signed though as we're not seeing any of the info log lines output: 
image

taylorsilva:
Closing because stale. Sorry folks. If anyone has bandwidth to work on this, feel free to re-submit the PR.

I wonder if the library support is any better for this now?
https://github.com/sigstore/cosign/tree/main/pkg/cosign
https://pkg.go.dev/github.com/sigstore/cosign/v2/pkg/cosign

PR #357, opened 2024-03-29
fix flakey check tests

taylorsilva:
At some point when adding these tests, we forgot that maps in Go do not guarantee that the order items are added is the order they'll appear in when iterating over the map: https://go.dev/blog/maps (see very last section)

This commit changes the `Tags` input into a slice, which does have a guaranteed order when iterating over.

Ran the tests locally many times and am no longer getting seeing any flakiness.


xtremerui:
thank you @taylorsilva !

Issue #356, opened 2024-03-28
images built with docker v25 or later fail to push


kalensk:
### Describe the bug

Images built with docker version 25 or later fail to push as shown in the below error.

For example, my task is pulling the latest docker in docker image (dind) to build my docker image. Using the latest dind tag which is docker version 26 fails, but using version 24-dind succeeds.
```
 - task: build_image
      privileged: true
      config:
        platform: linux
        image_resource:
          type: registry-image
          source:
            repository: docker
            tag: dind # tag: 24-dind succeeds
```

```
INFO[0000] pushing tag(s) dev                           
2024/03/28 01:42:37 existing blob: sha256:a7d483a98052b3c5108d6ed6d38caded5d2a787564ba0be1de08a5434f32a9db
2024/03/28 01:42:38 pushed blob: sha256:24c21885ea2bbe64a8a0b21901a61520772c747e98fb013d2cedf5bc23eecd4b
2024/03/28 01:42:39 pushed blob: sha256:617e845ae9918ce7686edff62e4c892c25ac172f75d75c76baef9abcbc8db052
2024/03/28 01:42:39 pushed blob: sha256:1249c494fb021137c410d273b1fbc3cfccf2daa4fb0c13ad223b31e052837235
2024/03/28 01:42:40 pushed blob: sha256:7ade3cec065838642366736d7d269cbefa2fea5eb839621bff0c520e00a18058
2024/03/28 01:42:41 pushed blob: sha256:052ad86d9dbd8a53393b7700d75dbd25764e564663bcc1a763341e3ac8b1c2af

[...]

PR #367, opened 2024-12-13
Bump `google/go-containerregistry` module version

HeavyWombat:
With an older version of the `google/go-containerregistry`, there seems to be an issue with pushing the image to the container registry when the image is built with newer Docker build engines.

Could be related to: https://github.com/concourse/registry-image-resource/issues/356.

What I tried:
- use `docker build` to create an image
  ```
  Server: Docker Engine - Community
  Engine:
    Version:          27.3.1
    API version:      1.47 (minimum version 1.24)
    Go version:       go1.22.7
    Git commit:       41ca978
    Built:            Fri Sep 20 11:41:03 2024
    OS/Arch:          linux/amd64
    Experimental:     false
  ```
- try to push using registry-image resource, which fails with: `MANIFEST_BLOB_UNKNOWN: blob unknown to registry`
- create custom registry-image resource with updated `google/go-containerregistry` dependency
- re-run the same build and push with custom resource, runs through successful


taylorsilva:
I'm bumping this and other deps. Thanks for raising this though!

PR #368, opened 2025-01-16
fix: support OCI Image Layout Specification

ae-govau:
This adds "oci-layout" as a format during "get" which saves out the contents per the OCI Image Layout Specification.

It also fixes the corresponding "put" to upload all images (including multi-architecture images) that are written according to this specification.

This handles both modern image indexes, as well as legacy image references.

Amongst other things, this makes this resource type more suitable for mirroring images from one repo to another, as this now will (a) preserve the digest and (b) mirror all images for each architecture in the source image.

### DIFF
--- a/README.md
+++ b/README.md
@@ -465,7 +465,7 @@ Fetches an image at the exact digest specified by the version.
 <tbody>
   <tr>
     <td><code>format</code> <em>(Optional)<br>Default: <code>rootfs</code></em></td>
-    <td>The format to fetch the image as. Accepted values are: <code>rootfs</code>, <code>oci</code></td>
+    <td>The format to fetch the image as. Accepted values are: <code>rootfs</code>, <code>oci</code>, <code>oci-layout</code></td>
   </tr>
   <tr>
     <td><code>skip_download</code> <em>(Optional)<br>Default: false</em></td>
@@ -486,7 +486,6 @@ The resource will produce the following files:
   For ECR images, this will include the registry the image was pulled from.
 * `./tag`: A file containing the tag from the version.
 * `./digest`: A file containing the digest from the version, e.g. `sha256:...`.
-* `./labels.json`: A file containing a JSON map of image labels, e.g. `{ "commit": "4e5c4ea" }`
 
 The remaining files depend on the configuration value for `format`:
 
@@ -501,16 +500,35 @@ In this format, the resource will produce the following files:
 
 * `./rootfs/...`: the unpacked rootfs produced by the image.
 * `./metadata.json`: the runtime information to propagate to Concourse.
+* `./labels.json`: A file containing a JSON map of image labels, e.g. `{ "commit": "4e5c4ea" }`
 
 ##### `oci` Format
 
-The `oci` format will fetch the image and write it to disk in OCI format. This
-is analogous to running `docker save`.
+The `oci` format will fetch the image and write it to disk in a format similar
+to running `docker save`.
 
 In this format, the resource will produce the following files:
 
 * `./image.tar`: the OCI image tarball, suitable for passing to `docker load`.
+* `./labels.json`: A file containing a JSON map of image labels, e.g. `{ "commit": "4e5c4ea" }`
 
+##### `oci-layout` Format
+
+The `oci-layout` format will fetch the image (or images) and write it to disk according to the
+[OCI Image Layout Specification](https://github.com/opencontainers/image-spec/blob/main/image-layout.md).
+
+In this format, the resource will produce the following files:
+
+* `./oci/index.json`
+* `./oci/oci-layout`
+* `./oci/blobs/sha256/aabbccdd...`
+* `./oci/blobs/sha256/ffeeddcc...`
+* `./oci/single-image-digest`: this is written only when the original digest specifies a legacy image,
+  rather than an image index. It is in the format `sha256:xxx`
+
+This format supports images with builds for different architectures, and is suitable for a corresponding
+`put` step, where the resultant put image will have the same digest as the one originally fetched
+(useful for mirroring use-cases).
 
 ### `put` Step (`out` script): push and tag an image
 
--- a/commands/in.go
+++ b/commands/in.go
@@ -135,6 +135,31 @@ func downloadWithRetry(tag name.Tag, source resource.Source, params resource.Get
 			return err
 		}
 
+		// handle oci-layout case first
+		if params.Format() == OciLayoutFormatName {
+			// first fetch the manifest
+			remoteDesc, err := remote.Get(repo.Digest(version.Digest), opts...)
+			if err != nil {
+				return fmt.Errorf("remote get: %w", err)
+			}
+
+			// wrap (as needed) as an index image
+			ioi, err := NewIndexImageFromRemote(remoteDesc)
+			if err != nil {
+				return fmt.Errorf("remote index or image: %w", err)
+			}
+
+			// write it out
+			err = ioi.WriteToPath(filepath.Join(dest, OciLayoutDirName))
+			if err != nil {
+				return fmt.Errorf("write oci layout: %w", err)
+			}
+
+			// and done
+			return nil
+		}
+
+		// else fallback to current behavior
 		image, err := remote.Image(repo.Digest(version.Digest), opts...)
 		if err != nil {
 			return fmt.Errorf("get image: %w", err)
--- a/commands/oci_layout.go
+++ b/commands/oci_layout.go
@@ -1,0 +1,235 @@
+package commands
+
+import (
+	"errors"
+	"fmt"
+	"io/fs"
+	"os"
+	"path/filepath"
+
+	v1 "github.com/google/go-containerregistry/pkg/v1"
+	"github.com/google/go-containerregistry/pkg/v1/empty"
+	"github.com/google/go-containerregistry/pkg/v1/layout"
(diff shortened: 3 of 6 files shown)


ae-govau:
The current output format of `oci` doesn't actually correspond to any published OCI Specification that I can find. However for this PR I added a 3rd output format `oci-layout` for `get` which does match the [OCI Image Layout Specification](https://github.com/opencontainers/image-spec/blob/main/image-layout.md).

While this should result in no changes to `get` (unless opted into with `format: oci-layout`), this PR does change the (currenlyt undocumented) behaviour of `put` when it refers to a directory.

Previously it would recognise an image in the `oci-layout` format, and would try to push it, but would fail if more than one image was present (such as multi-archictecture builds). With this change it should now succeed. Also previously it would confusing pick an odd digest for the image (it would find the first child manifest and use that, rather than the parent) - now it will correctly `put` the same digest that it previously `get`s. This PR also (if content trust is enabled) tries to submit each image found. I'm not sure if that is the best outcome, but I couldn't see a library method that would accept an ImageIndex so that seemed simplest.

I can see a build failure in this PR, however I'm unable to sign into your Concourse to check the build status to see why - tests did pass locally (including ones that push to repos) before submitting, however I didn't run the private repo tests.

ae-govau:
Looks like tests have passed now. Perhaps someone re-ran them?

taylorsilva (approved):
Thanks for the PR! This is a great feature to add to the resource. Sorry it took so long for me to review it.

taylorsilva:
@ae-govau Question for you, as I missed this during review and the ARM release work is having me dig into this more now.

Your comment here mentions a special marker file for legacy images: https://github.com/concourse/registry-image-resource/blob/fd13655dc02da42112c5d0dd468df400d89a4c0c/commands/oci_layout.go#L64-L67

I've been trying to find a reference to this practice _somewhere_ in either the OCI spec or in Docker's docs, but I've been unable to find anything about this. Searching for "single-image-digest" only yields this repo as the only result when I put it into Google. Searching Github returns the same thing, only our repo: https://github.com/search?q=%22single-image-digest%22&type=code

If this happens to not be a real thing then I'm planning to remove the related logic. No harm done, it's just looking like dead code to me. Let me know if I'm wrong please 🙏

ae-govau:
> @ae-govau Question for you, as I missed this during review and the ARM release work is having me dig into this more now.
> 
> Your comment here mentions a special marker file for legacy images:
> 
> https://github.com/concourse/registry-image-resource/blob/fd13655dc02da42112c5d0dd468df400d89a4c0c/commands/oci_layout.go#L64-L67
> 
> I've been trying to find a reference to this practice _somewhere_ in either the OCI spec or in Docker's docs, but I've been unable to find anything about this. Searching for "single-image-digest" only yields this repo as the only result when I put it into Google. Searching Github returns the same thing, only our repo: https://github.com/search?q=%22single-image-digest%22&type=code
> 
> If this happens to not be a real thing then I'm planning to remove the related logic. No harm done, it's just looking like dead code to me. Let me know if I'm wrong please 🙏 I'm working on this area of the code which is why I'm asking this question at all.

Hi @taylorsilva - yes I was expecting a question about that at some point. You are right - that particular name is not any particular spec.

The goal for us with this PR was to be to able to mirror images from one source (generally on the internet) to a private repo, and keep the same digest.

That meant that we needed to be able to handle both modern images (which may have more than one architecture) and the "legacy" ones.

The modern images include an index which refererences each underlying image, and are in the `oci-layout` format .

For legacy images (ie single images) we wrap them by creating an index file that references that images, then write it out as `oci-layout` format.

However we include that additional metadata so that a following `put` step can detect the the original image was a legacy image, and thus mirror it in the same format. The OCI Layout specification is clear that additional files (like this) are fine to include. Interestingly (when looking at the commit that added that clarification) the use-case they mention is very similar to ours - and possibly we could have achieved the same result that way:
https://github.com/opencontainers/image-spec/commit/60acaacb8003bff5882298333a4a017d6aedcd6a

If we didn't do that, then the `put` would result in an image with a new digest, and break our goal of mirroring into the private repo with the same digest.

For an example of a legacy image format, I'm pretty sure the output of https://github.com/concourse/oci-build-task are tarballs that end up in the legacy format when uploaded.

This code which writes out the directory probably has the better comment on it:
https://github.com/concourse/registry-image-resource/pull/368/files#diff-7f3cea05a6d2395d504681774b3fa7c576976893c1c2fbdd2902cd330838d7baR142-R165

So it's not dead code, even if it's only used in this repo. The `put` step uses it after its fetched with the `get` step.

I'm not fussed if we change the mechanism by which we handle legacy images, but I am keen to keep the property that ImageIndex and Image upstream types can be mirrored, preserving the digest, without the pipeline needing to know what format the image is in.

Hope that helps explain.

taylorsilva:
Thank you for that explanation @ae-govau.

The reason I ended up down this rabbit-hole is because I noticed that when pushing a multi-arch image made by any buildkit based tool (like the oci-build-task and Docker), buildkit would wrap the the images in an ImageIndex. So at the top-level I'd see an ImageIndex like this:

```json
{
  "schemaVersion": 2,
  "mediaType": "application/vnd.oci.image.index.v1+json",
  "manifests": [
    {
      "mediaType": "application/vnd.oci.image.index.v1+json",
      "digest": "sha256:55bf56746c776b4cc84498657dbaeeb33b247956dc08c6763b6ce2a7b53d2da2",
      "size": 647,
      "annotations": {
        "org.opencontainers.image.created": "2025-04-22T21:01:52Z"
      }
    }
  ]
}
```

The ImageIndex this references looks like this:
```json
{
  "schemaVersion": 2,
  "mediaType": "application/vnd.oci.image.index.v1+json",
  "manifests": [
    {
      "mediaType": "application/vnd.oci.image.manifest.v1+json",
      "digest": "sha256:4fce054b00b7a0363412c153bfec7e2093fd5dbd3a64256f21c1843bbc35f98e",
      "size": 669,
      "platform": {
        "architecture": "amd64",
        "os": "linux"
      }
    },
    {
      "mediaType": "application/vnd.oci.image.manifest.v1+json",
      "digest": "sha256:74b79ed32d2ba388471560515c3be5980c89d5143ae760a7b9a005e2b925907c",
      "size": 668,
      "platform": {
        "architecture": "arm64",
        "os": "linux"
      }
    },
    {
      "mediaType": "application/vnd.oci.image.manifest.v1+json",
      "digest": "sha256:d98cc734fea9bb1be4e1ae2fad74a34431cbd15fab44df1c99417e03b712a2a2",
      "size": 566,
      "annotations": {
        "vnd.docker.reference.digest": "sha256:4fce054b00b7a0363412c153bfec7e2093fd5dbd3a64256f21c1843bbc35f98e",
        "vnd.docker.reference.type": "attestation-manifest"
      },
      "platform": {
        "architecture": "unknown",
        "os": "unknown"
      }
    },
    {
      "mediaType": "application/vnd.oci.image.manifest.v1+json",
      "digest": "sha256:ba54321fca5fe50be961f8031aaca19a025045c8eacabeea12116d4528509044",
      "size": 566,
      "annotations": {
        "vnd.docker.reference.digest": "sha256:74b79ed32d2ba388471560515c3be5980c89d5143ae760a7b9a005e2b925907c",
        "vnd.docker.reference.type": "attestation-manifest"
      },
      "platform": {
        "architecture": "unknown",
        "os": "unknown"
      }
    }
  ]
}
```

If I did a normal build and push workflow with Docker I found that the second ImageIndex is what would get pushed. This is different from what the registry-image ends up doing. It ends up pushing the first ImageIndex which just holds a reference to the second ImageIndex. In Docker Hub this ends up masking the underlying images so you can't tell that the image you pushed is actually a multi-arch image:

image
(first image is pushed by registry-image, second was pushed using the oras CLI).

```
$> docker manifest inspect taylorsilva/file:latest
unsupported manifest type: *manifestlist.DeserializedManifestList

$> docker manifest inspect taylorsilva/file:oras
{
   "schemaVersion": 2,
   "mediaType": "application/vnd.oci.image.index.v1+json",
   "manifests": [
      {
         "mediaType": "application/vnd.oci.image.manifest.v1+json",
         "size": 669,
         "digest": "sha256:4fce054b00b7a0363412c153bfec7e2093fd5dbd3a64256f21c1843bbc35f98e",
         "platform": {
            "architecture": "amd64",
            "os": "linux"
         }
      },
      {
         "mediaType": "application/vnd.oci.image.manifest.v1+json",
         "size": 668,
         "digest": "sha256:74b79ed32d2ba388471560515c3be5980c89d5143ae760a7b9a005e2b925907c",
         "platform": {
            "architecture": "arm64",
            "os": "linux"
         }
      },
      {
         "mediaType": "application/vnd.oci.image.manifest.v1+json",
         "size": 566,
         "digest": "sha256:377a4b58f1fd97cbd0e3c2e402d3cb93c9c9bcbfa48f9d6555a9541cc31c43ae",
         "platform": {
            "architecture": "unknown",
            "os": "unknown"
         }
      },
      {
         "mediaType": "application/vnd.oci.image.manifest.v1+json",
         "size": 566,
         "digest": "sha256:aa349b319b267aea9da5e1210e4a321d0403b24540dc1ea104e40184167b1ef9",
         "platform": {
            "architecture": "unknown",
            "os": "unknown"
         }
      }
   ]
}
```


This is what lead me to this area of the code and wondering what was going on. I'm planning to update the behaviour of the registry-image to match what Docker and oras are doing.

PR #372, opened 2025-03-12
Update AWS SDK to v2

taylorsilva:
This PR does the following:

* Bumps all Go deps
* Bumps the module to Go 1.24
* Updates the AWS sdk to v2 since v1 is EOL this year

The `aws_ecr_registry_id` may be removed in a future version since AWS no longer requires you to specify the registry ID's in order to make cross-account ECR auth work.


taylorsilva:
We don't have any tests for the ECR auth. I'm going to manually test before merging.

taylorsilva:
Auth still works
image

PR #378, opened 2025-04-09
Fix multiple smaller issues in registry-image-resource

analytically:
- Moved regex compilation out of the loop in checkRepositoryRegex to prevent unnecessary recompilation on each iteration and added proper error handling
- Fixed an inconsistency in downloadWithRetry where OS.Stderr was used directly instead of the provided stderr parameter
- Corrected a typo in NewIndexImageFromRemote error message, changing "unspported" to "unsupported" for media type errors
- Fixed the repository handling in aliasesToBump, which was unnecessarily recreating the repository object despite already having a valid one passed as a parameter


taylorsilva (approved):
LGTM

PR #381, opened 2025-04-24
Unwrap nested Image Index's when pushing

taylorsilva:
I found that when pushing a multi-arch image to Docker Hub, I wouldn't
see which platforms the image supported like I would when pushing a
multi-arch image with the Docker CLI.

After digging into the OCI-layout I found out that buildkit, when
exporting the image to disk in oci-layout format, wraps the image index
in another image index. I'm not sure why it does that; haven't bothered
to dig into it in the buildkit code. It seems buildkit only does this
when writing the image to disk. My guess is it does it for legacy reason.
The ImageIndex wrapped looks like this:

```json
{
  "schemaVersion": 2,
  "mediaType": "application/vnd.oci.image.index.v1+json",
  "manifests": [
    {
      "mediaType": "application/vnd.oci.image.index.v1+json",
      "digest": "sha256:55bf56746c776b4cc84498657dbaeeb33b247956dc08c6763b6ce2a7b53d2da2",
      "size": 647,
      "annotations": {
        "org.opencontainers.image.created": "2025-04-22T21:01:52Z"
      }
    }
  ]
}
```

The problem is that when you push this wrapped ImageIndex, image
registries like Docker Hub don't know or display the platforms the image
supports.

When pushing the image using `docker push` or using the oras CLI to
push the local oci-layout image I made from buildkit, I found they both
pushed the nested image index instead of the parent. The nested image
index looks like this btw:

```json
{
  "schemaVersion": 2,
  "mediaType": "application/vnd.oci.image.index.v1+json",
  "manifests": [
    {
      "mediaType": "application/vnd.oci.image.manifest.v1+json",
      "digest": "sha256:4fce054b00b7a0363412c153bfec7e2093fd5dbd3a64256f21c1843bbc35f98e",
      "size": 669,
      "platform": {
        "architecture": "amd64",
        "os": "linux"
      }
    },
    {
      "mediaType": "application/vnd.oci.image.manifest.v1+json",
      "digest": "sha256:74b79ed32d2ba388471560515c3be5980c89d5143ae760a7b9a005e2b925907c",
      "size": 668,
      "platform": {
        "architecture": "arm64",
        "os": "linux"
      }
    }
  ]
}
```

This PR has the registry image unwrap all nested image indexes until
it gets an ImageIndex that contains one or more manifests that are not
another ImageIndex.

I've tested this and pushing multi-arch images with the registry-image
now correctly displays the platforms that were pushed. When mirroring
images, the original digests are still preserved as well.

I also did a little refactoring of the image pulling code. It got a bit hairy with the PR that added the oci-layout.

I also tried to see if it was possible to add progress bars while downloading the image. I found out that this is not supported by go-containerregistry: https://github.com/google/go-containerregistry/issues/670
The progress bars one sees when downloading in `rootfs` format is actually the progress of extracting the layers to disk. I've made the clear in the output now.

I think oras-go supports this since their CLI's show progress bars like docker: https://github.com/oras-project/oras-go
Switching to that would be a big overhaul though. It might be nicer to work with oras-go instead though.

### DIFF
--- a/commands/in.go
+++ b/commands/in.go
@@ -134,59 +134,59 @@ func downloadWithRetry(tag name.Tag, source resource.Source, params resource.Get
 			return err
 		}
 
-		// handle oci-layout case first
-		if params.Format() == OciLayoutFormatName {
-			// first fetch the manifest
-			remoteDesc, err := remote.Get(repo.Digest(version.Digest), opts...)
-			if err != nil {
-				return fmt.Errorf("remote get: %w", err)
-			}
-
-			// wrap (as needed) as an index image
-			ioi, err := NewIndexImageFromRemote(remoteDesc)
-			if err != nil {
-				return fmt.Errorf("remote index or image: %w", err)
-			}
-
-			// write it out
-			err = ioi.WriteToPath(filepath.Join(dest, OciLayoutDirName))
-			if err != nil {
-				return fmt.Errorf("write oci layout: %w", err)
-			}
-
-			// and done
-			return nil
+		switch params.Format() {
+		case "oci-layout":
+			return saveOciLayout(repo, version, dest, opts)
+		case "oci":
+			return saveOci(repo, tag, version, dest, opts)
+		case "rootfs":
+			return saveRootfs(repo, version, dest, opts, source.Debug, stderr)
+		default:
+			return fmt.Errorf("unknown format provided, must be one of 'rootfs', 'oci', 'oci-layout: %s", params.Format())
 		}
-
-		// else fallback to current behavior
-		image, err := remote.Image(repo.Digest(version.Digest), opts...)
-		if err != nil {
-			return fmt.Errorf("get image: %w", err)
-		}
-
-		err = saveImage(dest, tag, image, params.Format(), source.Debug, stderr)
-		if err != nil {
-			return fmt.Errorf("save image: %w", err)
-		}
-
-		return nil
 	})
 }
 
-func saveImage(dest string, tag name.Tag, image v1.Image, format string, debug bool, stderr io.Writer) error {
-	switch format {
-	case "oci":
-		err := ociFormat(dest, tag, image)
-		if err != nil {
-			return fmt.Errorf("write oci image: %w", err)
-		}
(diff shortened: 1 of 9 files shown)


mstandley-tempus:
@taylorsilva can you please cut a new concourse release that bundles v1.12.0 of registry-image instead of the v1.11.1 that comes with https://github.com/concourse/concourse/releases/tag/v7.13.1?  the latest version of concourse has a regression breaking multi-platform images, and it looks like this is the fix

taylorsilva:
@mstandley-tempus A regression? Can you share more details? What error or incorrect behaviour are you seeing?

Regardless, I am planning to cut a new release for 7.13.x with the updated resource types. The S3 resource type also had some issues that I want to get out for users.

mstandley-tempus:
> @mstandley-tempus A regression? Can you share more details? What error or incorrect behaviour are you seeing?
> 
> Regardless, I am planning to cut a new release for 7.13.x with the updated resource types. The S3 resource type also had some issues that I want to get out for users.

The regression is exactly what you've described in this PR: when we try to push a multi-platform docker image to the registry with v1.11.1, we get the nested index.  Trying to pull that image later gives an error like `no matching manifest for linux/amd64 in the manifest list entries`.  

Going back to registry-image v1.9.0 solves the problem.  We haven't tested v1.12.0 yet, but we're optimistic about it, because it contains the fix from this PR.

taylorsilva:
Okay cool, glad to know it fixed it for you. I never got that error using any of the prior versions. I'm a little sus'd by your error since it sounds related to pulling instead of pushing. Anyways, if it's working, it's working!

mstandley-tempus:
> Okay cool, glad to know it fixed it for you. I never got that error using any of the prior versions. I'm a little sus'd by your error since it sounds related to pulling instead of pushing. Anyways, if it's working, it's working!

Right, pushing the image up to the registry doesn't give us any kind of error.

We only see an error when we pull it down later, because the client (docker compose in this case) doesn't see the platform key in the top level manifest, and it's not smart enough to drill down to the nested manifest to find it.

Issue #49, opened 2019-07-25
Can't push OCI image layout


ikorolev93:
Reproduction:
```bash
#!/bin/bash
set -x
mkdir -p /tmp/repro
cd /tmp/repro
rm -f image.tar
buildah pull busybox
buildah push busybox oci-archive:image.tar
echo '{"params":{"image":"image/image.tar"},"source":{"repository":"example","tag":"latest"}}' \
  | podman run -it --rm -v $PWD/image.tar:/build/image/image.tar concourse/registry-image-resource /opt/resource/out /build
```

Output:
```
+ mkdir -p /tmp/repro
+ cd /tmp/repro
+ buildah pull busybox
db8ee88ad75f6bdc74663f4992a185e2722fa29573abcc1a19186cc5ec09dceb
+ rm -f image.tar
+ buildah push busybox oci-archive:image.tar
Getting image source signatures
Copying blob 0d315111b484 done
Copying config d068e40b11 done
Writing manifest to image destination
Storing signatures
Successfully pushed image.tar:@sha256:42212762065786db5a6fee614c801044c9328da5389b30c314c001d1e186a2ce
+ echo '{"params":{"image":"image/image.tar"},"source":{"repository":"example","tag":"latest"}}'
+ podman run -it --rm -v /tmp/repro/image.tar:/build/image/image.tar concourse/registry-image-resource /opt/resource/out /build
{"params":{"image":"image/image.tar"},"source":{"repository":"example","tag":"latest"}}
ERRO[0000] could not load image from path 'image/image.tar': file manifest.json not found in tar 
```

Indeed, there is no `manifest.json` in the [OCI Image Layout Specification](https://github.com/opencontainers/image-spec/blob/master/image-layout.md).

As a workaround, I changed my task to do `buildah push whatever docker-archive:image.tar`, but it would be nice if `registry-image-resource` supported OCI images.

cirocosta:
Thanks for reporting, @deadNightTiger !

I noticed that this is also true (for the same reason) for anyone leveraging `buildkit` directly with `--output type=oci,dest=image.tar`. 

(for anyone wanting to try that out, check https://github.com/cirocosta/buildkit-task).

vito:
@cirocosta Broken link - maybe the repo's private?

vito:
[...]

PR #383, opened 2025-05-22
Update description of image param format for put step in README.md

evgnblkn:
This PR adds instruction how to set `image` param in a put step when use `oci` image tarball. Additionaly for #368 #49


taylorsilva (approved):
Thanks for updating that!

PR #386, opened 2025-05-29
use wolfi as base image

taylorsilva:
sets up the image to be built for multiple platforms


taylorsilva:
Check pipeline broken right now, merging anyways.

PR #390, opened 2025-08-29
support zstd layer decompression

h0nIg:


commands/rootfs.go
```diff
@@ -78,12 +79,34 @@ func extractLayer(dest string, layer v1.Layer, bar *mpb.Bar, chown bool) error {
...
+		log.Debugf("failed to decompress layer with gzip, trying zstd instead")
+
+		zr, err := zstd.NewReader(bar.ProxyReader(r))
+		crc = func() error { zr.Close(); return nil }
+		cr = zr
+		if err != nil {
+			log.Errorf("failed to decompress layer with gzip / zstd")
+			return err
+		}
+	}
```
taylorsilva, line 107: Instead of blindly falling back to trying zstd, why not inspect the `layer.MediaType()` for `types.OCILayer`/`types.OCILayerZStd`? Will make it easier in the future as well if we need to support more compression methods.

taylorsilva (changes requested):
Thanks for the PR! No need to worry about tests here. I think we would need to push a new test image that's compressed with zstd before we write a test. I'll figure that out later.

taylorsilva:
@Kump3r no blockers and will definitely be in the next release.

Kump3r:
Based on my limited knowledge on the subject we will need to either extend the existing function for building in ([setup-images](https://github.com/concourse/registry-image-resource/blob/master/testdata/setup-images)), or create a new dedicated function. Those will require to call docker with `buildx` and `—compression=zstd` and `—oci-mediatypes` options, which will be an additional dependency for contributors imho. In any case we shouldn’t need a new image specifically and could reuse some of the existing. I am a bit preoccupied atm, but if the PR becomes stale, I can try allocating some time to get this in v8, as it should be hugely beneficial for our environment. Thanks for the PR!

Kump3r:
Seems like the “against a private repo with credentials” tests are failing. Could it be that a tocken for the registry expired or smth?

taylorsilva:
I think I just got the digests a bit mixed up. Fingers cross what I just pushed fixes it.

taylorsilva (approved):

PR #394, opened 2026-01-02
Capture image ENTRYPOINT and CMD values in rootfs format

taylorsilva:
When exporting an image into `rootfs` format, include `entrypoint` and `cmd` from the OCI image, if present, into the `metadata.json` file that sits at the same level as the `rootfs/` directory.

### DIFF
--- a/commands/in.go
+++ b/commands/in.go
@@ -18,8 +18,10 @@ import (
 )
 
 type ImageMetadata struct {
-	Env  []string `json:"env"`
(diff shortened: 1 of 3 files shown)


taylorsilva:
I tested to make sure this didn't break anything with the guardian runtime. I wasn't sure if guardian read the `metadata.json` file itself, but it appears not to, or at least doesn't care about any extra top-level keys we add.

PR #403, opened 2026-05-15
Add note and test about tags prefixed with v

jghiloni:


jghiloni:
obviously nothing here necessitates spinning up a new version

PR #405, opened 2026-06-24
Detect layer compression by content, not media type

mattpollock:
## What

`extractLayer` chose the decompressor from the layer's **declared media type**: only `types.OCILayerZStd` was treated as zstd, and everything else fell through to gzip. That breaks for zstd layers labeled with a non-OCI media type — most commonly **BuildKit pushing with Docker schema2**, which emits `application/vnd.docker.image.rootfs.diff.tar.zstd`. The media-type switch routes those to `gzip.NewReader`, so `in` fails with:

```
write rootfs: extract image: gzip: invalid header
```

This is increasingly common as people adopt zstd-layer base images (e.g. hardened/minimal bases) and push them through standard `docker/build-push-action` without forcing OCI media types.

## Fix

Detect the compression from the layer's own leading **magic bytes** instead of trusting the media-type label:

- zstd -> `28 b5 2f fd`
- gzip -> `1f 8b`
- otherwise -> read as an uncompressed tar

So gzip, zstd, and uncompressed layers all unpack regardless of whether they're labeled OCI or Docker media types. `mediaType` is kept in the signature for compatibility but no longer drives the decision.

## Repro

```sh
# zstd layers under Docker schema2 media types (oci-mediatypes=false)
docker buildx build --output type=image,name=$IMG,push=true,compression=zstd,force-compression=true,oci-mediatypes=false .
```
`get` of `$IMG` fails with `gzip: invalid header` before this change; succeeds after.

## Notes

- No behavior change for existing gzip / OCI-zstd images (same decompressors, just selected by content).
- Single file (`commands/rootfs.go`). Happy to add a regression test fixture (a Docker-schema2 zstd image alongside the existing OCI-zstd one in `testdata/setup-images`) if you'd like it in this PR.

### DIFF
--- a/commands/rootfs.go
+++ b/commands/rootfs.go
@@ -2,6 +2,8 @@ package commands
 
 import (
 	"archive/tar"
+	"bufio"
+	"bytes"
 	"compress/gzip"
 	"io"
 	"os"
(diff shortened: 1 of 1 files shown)


commands/rootfs.go
```diff
@@ -195,26 +197,53 @@ func extractLayer(dest string, layer v1.Layer, bar *mpb.Bar, chown bool) error {
...
+//
+// It detects the compression from the layer's own leading bytes rather than
+// trusting the declared media type. Builders and registries don't always label
+// zstd layers with the OCI media type (types.OCILayerZStd); BuildKit pushing
+// with Docker schema2 emits application/vnd.docker.image.rootfs.diff.tar.zstd,
+// which a media-type switch mis-routes to gzip and fails with
+// "gzip: invalid header". Sniffing the magic bytes handles gzip, zstd, and
+// uncompressed tar regardless of how the layer is labeled. mediaType is kept
+// for signature compatibility but no longer drives the decision.
 func compressionReader(layer io.ReadCloser, mediaType types.MediaType, bar *mpb.Bar) (*tar.Reader, compressReaderClose, error) {
```
taylorsilva, line 216: This func is private and only used in one place. Let's just update the signature and update where it's called, which should be higher up in this file.
```suggestion
// uncompressed tar regardless of how the layer is labeled. mediaType is kept
// for signature compatibility but no longer drives the decision.
func compressionReader(layer io.ReadCloser, bar *mpb.Bar) (*tar.Reader, compressReaderClose, error) {
```

And update here: https://github.com/concourse/registry-image-resource/blob/1627112f6e2314e5450de4a84f9b28aa087169ee/commands/rootfs.go#L83-L91

commands/rootfs.go
```diff
@@ -195,26 +197,53 @@ func extractLayer(dest string, layer v1.Layer, bar *mpb.Bar, chown bool) error {
 
 type compressReaderClose func() error
 
+// Layer compression magic numbers.
```
taylorsilva, line 200: nit: for easy lookup reference in the future link to https://en.wikipedia.org/wiki/List_of_file_signatures

```suggestion
// Layer compression magic numbers. Can find more at:
// https://en.wikipedia.org/wiki/List_of_file_signatures
```

taylorsilva (changes requested):
Thanks for the PR! This LGTM, just make one final clean up of the func signature and then I'll merge. How annoying is it to have standards and then NOT have them followed 😭 

P.S. Can you complete the EasyCLA too, please and TY 🙏

taylorsilva:
@mattpollock Friendly ping as a reminder about this PR. I'm working on getting the next Concourse release out and it would be nice to have this included. TY!

mattpollock:
I'm closing this PR on account of not being able to complete the CLA auth. Do with the content whatever you like.

PR #408, opened 2026-07-17
Add `annotation_regex` and `label_regex` filtering to in step metadata

IvanChalukov:
closes #398 

#### Changes introduced:
* Adds two new optional source fields - `annotation_regex` and `label_regex` - that filter OCI image manifest annotations and config labels (by key) into the `get` step metadata output
* Extracts metadata assembly into commands/metadata.go with a shared `filterMap` helper, eliminating duplication between annotation and label fetching

#### Details
Both `annotation_regex` and `label_regex` accept a Go regular expression matched against metadata keys. Only matching keys are included in the response. When omitted, no annotations or labels are added to the metadata.

Example source:
```yaml
 source:
    repository: ichalukov/my-app
    tag: lin1
    label_regex: "^org\\.opencontainers\\.image\\."
    annotation_regex: "t([a-z]+)le"
```

### DIFF
--- a/commands/metadata.go
+++ b/commands/metadata.go
@@ -1,0 +1,104 @@
+package commands
+
+import (
+	"fmt"
+	"regexp"
+	"sort"
+
+	resource "github.com/concourse/registry-image-resource"
+	v1 "github.com/google/go-containerregistry/pkg/v1"
+)
+
+func buildImageMetadata(source resource.Source, version resource.Version, image v1.Image) ([]resource.MetadataField, error) {
+	metadata := append(source.Metadata(), resource.MetadataField{
+		Name:  "tag",
+		Value: version.Tag,
+	})
+
+	fields, err := collectMetadataFields(source, image)
+	if err != nil {
+		return nil, err
+	}
+
+	return append(metadata, fields...), nil
+}
+
+func collectMetadataFields(
+	source resource.Source,
+	image v1.Image,
+) ([]resource.MetadataField, error) {
+	var fields []resource.MetadataField
+
+	if regex := source.AnnotationRegex; regex != "" {
+		annotations, err := fetchImageAnnotations(regex, image)
+		if err != nil {
+			return nil, fmt.Errorf("fetch annotations: %w", err)
+		}
+
+		fields = append(fields, metadataFieldsFromMap(annotations)...)
+	}
+
+	if regex := source.LabelRegex; regex != "" {
+		labels, err := fetchImageLabels(regex, image)
+		if err != nil {
+			return nil, fmt.Errorf("fetch labels: %w", err)
+		}
+
+		fields = append(fields, metadataFieldsFromMap(labels)...)
+	}
+
+	return fields, nil
+}
+
+func fetchImageAnnotations(regex string, img v1.Image) (map[string]string, error) {
+	manifest, err := img.Manifest()
+	if err != nil {
+		return nil, fmt.Errorf("get manifest: %w", err)
+	}
+
+	return filterMap(regex, manifest.Annotations, "annotation")
+}
+
+func fetchImageLabels(regex string, img v1.Image) (map[string]string, error) {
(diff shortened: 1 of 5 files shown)


commands/metadata.go
```diff
@@ -0,0 +1,127 @@
...
+	platform := source.Platform(params.RawPlatform)
+	opts = append(opts, remote.WithPlatform(v1.Platform{
+		Architecture: platform.Architecture,
+		OS:           platform.OS,
+	}))
+
+	image, err := remote.Image(repo.Digest(version.Digest), opts...)
+	if err != nil {
+		return nil, fmt.Errorf("get image: %w", err)
+	}
```
taylorsilva: First thought I had here was: don't we make similar API calls beforehand when downloading the image? We'd be making an extra API call for information that we might have previously gotten already when we downloaded the image. I try to think about how our resources can be good API consumers, which means making the least number of calls necessary.

taylorsilva: Okay, so depending on the format we save the image in we do call `remote.Get` or `remote.Image`:
https://github.com/concourse/registry-image-resource/blob/b77ed381ed7c083b868db8c516853a61b9d5e4a3/commands/in.go#L169-L170
https://github.com/concourse/registry-image-resource/blob/b77ed381ed7c083b868db8c516853a61b9d5e4a3/commands/in.go#L188-L189
https://github.com/concourse/registry-image-resource/blob/b77ed381ed7c083b868db8c516853a61b9d5e4a3/commands/in.go#L200-L201

The last two have a `v1.Image` object that we could re-use and feed into the metadata functions you've added. `remote.Get` can get us to a `v1.image` object as well by calling `.Image()` on the `Descriptor` it returns.

This would be a bit of a pain to refactor. Had a few ideas how it could be approached and current leading thought is to pull out the `remote.Get` and `remote.Image` to before this if-statement here: https://github.com/concourse/registry-image-resource/blob/b77ed381ed7c083b868db8c516853a61b9d5e4a3/commands/in.go#L88

Do `remote.Get` before that `if` and pass the `v1.Descriptor` into the `downloadWithRetry()` function and then into the `save***()` functions.

Both `remote.Get/Image` download the image manifest, so I think it's fine if we don't have that wrapped in the `RetryOnTransientError()` function. The errors that function catches happen during the downloading of the image layers, not when the manifest is being downloaded, which is relatively small compared to the layers.

Are you open to giving this refactor a try so we can avoid making a second API call for the image manifest?

IvanChalukov: Hi @taylorsilva, thanks for pointing this out. I had considered the extra API call concern as well, but I wasn't sure whether changing the existing download flow was in scope for this PR.

I'm happy to refactor this. Passing the fetched image/descriptor through the existing flow seems like the right direction, especially since it lets us reuse data we already retrieved from the registry. I'll explore moving the fetch earlier and adjusting the save* functions accordingly.

IvanChalukov: Hi @taylorsilva. Sorry for late update. I wanted to play around and understand the whole implementation part around skip_dowload option.

I considered reusing the fetched image for metadata, and I agree that avoiding duplicate registry calls is important.

However, I think `skip_download` should be treated as a stronger contract: when it is set, the resource should not make any registry calls. The option is described as "Skip downloading the image", and the main use cases are triggering jobs without needing the image info(no calls to registry) or after a put step where the image is already available.

Because of that, I don't think we should fetch the image only to populate labels/annotations when `skip_download` is enabled. That would make `skip_download` surprising and could still fail pipelines due to registry issues like authentication errors or 429 responses.

The new behavior is:

- skip_download=false: download the image and reuse the same v1.Image instance for metadata extraction.
- skip_download=true: do not contact the registry and only return base metadata.

I updated the tests to reflect this behavior. If we want metadata fetching without downloading in the future, I think that should be an explicit opt-in feature rather than being implied by label_regex/annotation_regex.

Curious to hear your thoughts on this - does this approach make sense to you?

